PR #5, opened 2018-03-16
SLING-7544 - improving optimized alias lookup to not block during int…

DominikSuess:
…ialization but fall back to default and enforce non traversal query for building up aliasmap

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -25,6 +25,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -111,6 +112,10 @@ public class MapEntries implements
     public static final int DEFAULT_DEFAULT_VANITY_PATH_REDIRECT_STATUS = HttpServletResponse.SC_FOUND;
 
     private static final String JCR_SYSTEM_PREFIX = "/jcr:system/";
+
+    final static String ALIAS_QUERY_DEFAULT = "SELECT sling:alias FROM nt:base WHERE sling:alias IS NOT NULL";
+
+    final static String ALIAS_QUERY_NO_TRAVERSAL = ALIAS_QUERY_DEFAULT + " option(traversal fail)";
 
     static final String ANY_SCHEME_HOST = "[^/]+/[^/]+";
 
@@ -1052,10 +1057,7 @@ public class MapEntries implements
      */
     private Map<String, Map<String, String>> loadAliases(final ResourceResolver resolver) {
         final Map<String, Map<String, String>> map = new ConcurrentHashMap<>();
-        String queryString = "SELECT sling:alias FROM nt:base WHERE sling:alias IS NOT NULL";
-		if (this.factory.isForceNoAliasTraversal()) {
-			queryString += " option(traversal fail)";
-		}
+		String queryString = this.factory.isForceNoAliasTraversal() ? ALIAS_QUERY_NO_TRAVERSAL : ALIAS_QUERY_DEFAULT;
 		while (true){
 	        try {
 		        final Iterator<Resource> i = resolver.findResources(queryString, "sql");
@@ -1064,26 +1066,37 @@ public class MapEntries implements
 		            loadAlias(resource, map);
 		        }
 		        break;
-	    	} catch (QuerySyntaxException e) {
-			Throwable cause = unwrapThrowable(e);
-			if (cause instanceof IllegalArgumentException) {
-				log.debug(
-						"Expected index not available or traversal fail option not known and traversal is prevented - will retry",
-						e);
-				try {
-					TimeUnit.MILLISECONDS.sleep(getTraversalRetryInterval());
-				} catch (InterruptedException ex) {
-					log.warn("Interrupted while sleeping", ex);
+	    	} catch (SlingException e) {
+	    		Throwable cause = unwrapThrowable(e);
+	    		if (cause instanceof IllegalArgumentException) {
+					log.debug(
+						"Expected index not available yet - will retry", e);
+					try {
+						TimeUnit.MILLISECONDS.sleep(getTraversalRetryInterval());
+					} catch (InterruptedException ex) {
+						log.warn("Interrupted while sleeping", ex);
+					}
+				} else if (cause instanceof ParseException) {
+	    			if (ALIAS_QUERY_NO_TRAVERSAL.equals(queryString)) {
+						log.warn("Traversal fail option set but query not accepted by queryengine, falling back to allowing traversal as queryengine might not support option", e);
+						queryString = ALIAS_QUERY_DEFAULT;
+					} else {
+						log.error("Queryengine couldn't parse query - interrupting loading of aliasmap",e);
+						break;
+					}
+					try {
+						TimeUnit.MILLISECONDS.sleep(getTraversalRetryInterval());
+					} catch (InterruptedException ex) {
+						log.warn("Interrupted while sleeping", ex);
+					}
+	    		
+					
+				} else {
+					log.error("QueryEngine not able to process query {} ", queryString, e);
+					break;
 				}
-			} else {
-				log.error("QueryEngine not able to process query {} ", queryString, e);
-				break;
-			}
-
-		} catch (SlingException ex) {
-	        	log.error("Loading of aliases via query failed.", ex);
-	        }
-		}   
+	    	}   
+		}
         return map;
     }
 
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -133,9 +133,9 @@ public class MapEntries implements
 
     private Map <String,List <String>> vanityTargets;
 
-    private Map<String, Map<String, String>> aliasMap;
+    private volatile Map<String, Map<String, String>> aliasMap;
     
-	private boolean isAliasMapInitialized = false;
+	private volatile boolean isAliasMapInitialized = false;
 
     private final ReentrantLock initializing = new ReentrantLock();
 
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -1088,9 +1088,9 @@ public class MapEntries implements
     }
 
     /**
-     * ex
-     * @param e
-     * @return
+     * Extract root cause of exception
+     * @param e {@code Throwable} to be checked
+     * @return Root {@code Throwable} 
      */
     private Throwable unwrapThrowable(Throwable e) {
 		return e.getCause() == null ? e : unwrapThrowable(e.getCause());
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -78,6 +78,7 @@ public class MapEntries implements
     ExternalResourceChangeListener {
 
     private static final String JCR_CONTENT = "jcr:content";
+    private static final int TRAVERSAL_RETRY_TIMER = 30000;
 
     private static final String JCR_CONTENT_PREFIX = "jcr:content/";
 
@@ -191,19 +192,20 @@ public class MapEntries implements
 
             final Map<String, List<MapEntry>> newResolveMapsMap = new ConcurrentHashMap<>();
 
+            isAliasMapInitialized = false;
             //optimization made in SLING-2521
             if (this.factory.isOptimizeAliasResolutionEnabled()) {
+    			new Thread(new Runnable(){
+    				public void run() {
+    					aliasMap = loadAliases(resolver);
+    					isAliasMapInitialized = true;
+    					}
+    				}).start();
             }
 
             this.resolveMapsMap = newResolveMapsMap;
-            isAliasMapInitialized = false;
+
             doUpdateConfiguration();
-				new Thread(new Runnable(){
-					public void run() {
-						aliasMap= loadAliases(resolver);
-						isAliasMapInitialized = true;
-						}
-					}).start();
 
             sendChangeEvent();
         } catch (final Exception e) {
@@ -266,6 +268,10 @@ public class MapEntries implements
             this.initializing.unlock();
         }
 
+    }
+    
+    protected int getTraversalTimer(){
+    	return TRAVERSAL_RETRY_TIMER;
     }
 
     private boolean addResource(final String path, final AtomicBoolean resolverRefreshed) {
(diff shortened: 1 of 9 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -127,7 +128,9 @@
 
     private Map <String,List <String>> vanityTargets;
 
-    private Map<String, Map<String, String>> aliasMap;
+    private volatile Map<String, Map<String, String>> aliasMap;
```
rombert: Two volatile flags don't ensure consistency. Either use a synchronized block and non-volatile fileds or ensure thread safety in some other manner.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -189,10 +192,15 @@ protected void doInit() {
 
             final Map<String, List<MapEntry>> newResolveMapsMap = new ConcurrentHashMap<>();
 
+            isAliasMapInitialized = false;
             //optimization made in SLING-2521
             if (this.factory.isOptimizeAliasResolutionEnabled()) {
-                final Map<String, Map<String, String>> aliasMap = this.loadAliases(resolver);
-                this.aliasMap = aliasMap;
+    			new Thread(new Runnable(){
```
rombert: What happens when the component is stopped/started quickly in succession? How do you ensure that no leftover threads are left running?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1027,12 +1044,28 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
         final Map<String, Map<String, String>> map = new ConcurrentHashMap<>();
-        final String queryString = "SELECT sling:alias FROM nt:base WHERE sling:alias IS NOT NULL";
-        final Iterator<Resource> i = resolver.findResources(queryString, "sql");
-        while (i.hasNext()) {
-            final Resource resource = i.next();
-            loadAlias(resource, map);
-        }
+        String queryString = "SELECT sling:alias FROM nt:base WHERE sling:alias IS NOT NULL";
+		if (this.factory.isForceNoAliasTraversal()) {
+			queryString += " option(traversal fail)";
```
rombert: Does this introduce a dependency on a certain version of Oak?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -261,6 +269,10 @@ protected void initializeVanityPaths() throws IOException {
         }
 
     }
+    
+    protected int getTraversalTimer(){
```
rombert: 'timer' is maybe not the best term here. Delay? Interval?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1027,12 +1044,28 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
+		}
+		while (true){
+	        try {
+		        final Iterator<Resource> i = resolver.findResources(queryString, "sql");
+		        while (i.hasNext()) {
+		            final Resource resource = i.next();
+		            loadAlias(resource, map);
+		        }
+		        break;
+	        } catch (Exception e) {
```
rombert: Is `Exception` the right thing to catch here? I think we can use more fine-grained exceptions.  Also what happens to resolution calls when initialisation never completes?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1027,12 +1044,28 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
+	        	log.debug("Expected index not available and no traversal enforced", e);
+	            try {
+	                TimeUnit.MILLISECONDS.sleep(getTraversalTimer());
+	            } catch (InterruptedException ex) {
+	                log.warn("Interrupted while sleeping");
```
rombert: The root exception should also be logged. But this is executed in the thread started in MapEntries.init(), right? This means that a failure will not be propagated to the caller.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
```diff
@@ -168,6 +172,8 @@ public void test_simple_alias_support() {
         });
 
         mapEntries.doInit();
+        
+        while(!mapEntries.isAliasMapInitialized()){}
```
rombert: An empty loop looks wrong. I know what it does :-) but it looks wrong. I would suggest either adding a method to mapEntries for waiting for it to be available or using a test helper method.

Also please add timeouts to all methods having such async logic to prevent them from hanging indefinitely under special circumstances.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
DominikSuess: @rombert  thx - in fact the volatile is obsolete and was just a leftover of an attempt to fix an issue that turned out to be caused by the mocking flow - the map used already is a ConcurrentHashMap - so I'll drop the volatile marker

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
DominikSuess: @rombert  - the thread at the moment would afair simply complete - certainly quickly starting and stopping would let threads pile up - so I'll capture the thread reference and interrupt in stop() (everything else seems to be overly complicated for a rather virtual case).

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
DominikSuess: Yes this option was added in https://issues.apache.org/jira/browse/OAK-4888 - this is the reason why I made this configurable. Afair (to be confirmed) the query will fail with corresponding exception if the queryengine doesn't understand the option.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
DominikSuess: TraversalRetryInterval sounds good.. a bit mindless on picking name - will fix.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
DominikSuess: @rombert  - I did try to find an appropriate answer and honestly did some cargo culting on that one as the cases handling failing traversals that I could find did do it that way - I'll crosscheck the oak codebase if we have a reliable behavior - esp covering the case as well where the queryengine doesn't understand the option for more refined logmessage.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
DominikSuess: @rombert  - I'll be adding timeout to the Test annotation - this is really kind of a test hack  to wait for this to complete. I can't see any use case for waiting for the alias map in reality so I'm a bit hesitant to add logic to the MapEntries class just to trigger some waiting test. I'll be checking if there is some mockito helper to watch out for a certain condition to change.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
catholicon: Quick note as @DominikSuess pinged me offline. I wonder if it might an interesting idea to do a {{EXPLAIN SELECT sling:alias .... }} first and see that {{/\* travers}} doesn't show up in the plan.

That should work across oak versions afaict (I just tested with oak 1.2.2)

DominikSuess: @catholicon thx for the suggestion, it's a nice attempt but I assume it's a rather academical case and textoutput also is some kind of implementation detail that might more easily change - @rombert  I just fixed the exception handling for the non traverse option - I leave it up to you to answer the question if we should go for the more backward compatible option which requires text parsing or the option I provided which is "configurable".

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -128,9 +133,9 @@
 
     private Map <String,List <String>> vanityTargets;
 
-    private volatile Map<String, Map<String, String>> aliasMap;
+    private Map<String, Map<String, String>> aliasMap;
     
-	private volatile boolean isAliasMapInitialized = false;
+	private boolean isAliasMapInitialized = false;
```
rombert: Sorry for the late reply, I missed your additional changes.

Right now the situation is no better than in the original submission. You're reading/writing to `aliasMap` and `isAliasMapInitialized` from different threads and there is no guarantee that the changes will be visible - no synchronized/volatile/etc.

The aliasMap may be a concurrent implementation, but the reference setting is not thread-safe. The solution you choose should at least:

- ensure that aliasMap/isAliasMapInitialized values are changed atomically
- access to aliasMap/isAliasMapInitialized is done in a thread-safe manner

DominikSuess: @rombert  ok I revisited the scenario and added the volatile markers again. This should now be sufficient as access is now threadsafe - the aliasmap is not manipulated in the thread despite of assigning a new map I don't see another multi-threading issue.

rombert: The changes you made don't ensure that the values are changed atomically. Assuming that we have thread `T1` updating the values and thread `T2` reading the values it's possible for thread `T1` to set value 1, then `T2` to read value 1 (new) and value 2 (old).

You should ensure that the values are changed atomically, for instance by using synchronized blocks or by using a single volatile field that holds both the aliasMap data and the init ready flag.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -189,10 +199,16 @@ protected void doInit() {
...
             if (this.factory.isOptimizeAliasResolutionEnabled()) {
-                final Map<String, Map<String, String>> aliasMap = this.loadAliases(resolver);
-                this.aliasMap = aliasMap;
+            	aliasTraversal = new Thread(new Runnable(){
+    				public void run() {
+    					aliasMap = loadAliases(resolver);
+    					isAliasMapInitialized = true;
+    					}
+    				});
+            	aliasTraversal.start();
```
rombert: Are we missing an `else` block here which loads the aliases in the main thread?

rombert:
@DominikSuess  - I think we can solve the concurrency issue with the following patch:

```
diff --git a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
index 6cc51e7..03ddab4 100644
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -25,7 +25,6 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -50,7 +49,6 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReentrantLock;
 
-import javax.jcr.query.InvalidQueryException;
 import javax.servlet.http.HttpServletResponse;
 
 import org.apache.sling.api.SlingConstants;
@@ -134,8 +132,6 @@ public class MapEntries implements
     private Map <String,List <String>> vanityTargets;
 
     private volatile Map<String, Map<String, String>> aliasMap;
-    
-	private volatile boolean isAliasMapInitialized = false;
 
     private final ReentrantLock initializing = new ReentrantLock();
 
@@ -162,7 +158,7 @@ public class MapEntries implements
         this.resolveMapsMap = Collections.singletonMap(GLOBAL_LIST_KEY, (List<MapEntry>)Collections.EMPTY_LIST);
         this.mapMaps = Collections.<MapEntry> emptyList();
         this.vanityTargets = Collections.<String,List <String>>emptyMap();
-        this.aliasMap = Collections.<String, Map<String, String>>emptyMap();
+        this.aliasMap = Collections.emptyMap();
 
         doInit();
 
@@ -199,13 +195,12 @@ public class MapEntries implements
 
             final Map<String, List<MapEntry>> newResolveMapsMap = new ConcurrentHashMap<>();
 
-            isAliasMapInitialized = false;
+            aliasMap = Collections.emptyMap();
             //optimization made in SLING-2521
             if (this.factory.isOptimizeAliasResolutionEnabled()) {
             	aliasTraversal = new Thread(new Runnable(){
     				public void run() {
     					aliasMap = loadAliases(resolver);
-    					isAliasMapInitialized = true;
     					}
     				});
             	aliasTraversal.start();
@@ -656,7 +651,12 @@ public class MapEntries implements
 
 	@Override
 	public boolean isAliasMapInitialized() {
-		return isAliasMapInitialized;
+	    // since loading the aliases is equivalent to replacing the empty map
+	    // with another instance, and the reference is volatile, it's safe
+	    // to equate initialization being done with the empty map being replaced
+	    // note that it's not provably safe to check the map size, as we might
+	    // enter the scenario where there are no aliases
+		return aliasMap != Collections.<String,Map<String,String>> emptyMap();
 	}
 
     /**
```

Feel free to apply this to your PR, but also make sure to check the rest of the review comments.

DominikSuess:
@rombert  - thanks, I incorporated as proposed as this certainly eliminates also theoretical cases. The remaining question was about oak version specifics. The only improvement to given scenario I could imagine is, that we "assume" the capability is given and switch that feature off in case the query leads to the corresponding parse exception that oak would issue. //cc @catholicon  for comment

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1027,16 +1052,51 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
+	        try {
+		        final Iterator<Resource> i = resolver.findResources(queryString, "sql");
+		        while (i.hasNext()) {
+		            final Resource resource = i.next();
+		            loadAlias(resource, map);
+		        }
+		        break;
+	    	} catch (QuerySyntaxException e) {
+			Throwable cause = unwrapThrowable(e);
+			if (cause instanceof IllegalArgumentException) {
```
catholicon: Maybe, also see that fail for traversal is set!?.... I don't see why that should be the case, but being a bit defensive won't hurt I guess...

DominikSuess:
@catholicon , @rombert  added corresponding logic to warn and automatically fall back to the query without traversal fail in case engine not capable of parsing the option (crosschecked with older versions - lowest cause always was ParseException  before the options got introduced).

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
DominikSuess: @catholicon  ok I constrained it to check for the right query to be used so the retry behavior will only kick in when the corresponding query is active.

rombert:
@DominikSuess - there's still an open question at https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/5/files/51a0d36c39f1a59bbb99b8337ac99e727f8a6654#r184093639 . Can you please answer?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
DominikSuess: @rombert  - the patch only affects the if block - in the other case none of the query based caching happens and nothing needs to be loaded.

rombert: Ah right. Thanks!

rombert:
Applied, thanks for the contribution @DominikSuess !

PR #7, opened 2018-05-04
SLING-7544 - fixing potential race between intialization and addition…

DominikSuess:
… of updated aliases

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -137,8 +137,12 @@ public class MapEntries implements
     private Map <String,List <String>> vanityTargets;
 
     private volatile Map<String, Map<String, String>> aliasMap;
+    
+    private volatile Map<String, Map<String, String>> aliasQueueMap;
 
     private final ReentrantLock initializing = new ReentrantLock();
+    
+    private final ReentrantLock aliasMerging = new ReentrantLock();
 
     private final AtomicLong vanityCounter;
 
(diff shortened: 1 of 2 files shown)


rombert:
Thanks for following-up @DominikSuess  . Can you please explain what the race condition is/was?

DominikSuess:
@rombert initialization of aliasmap(esp. iteration over resultset) might race against further changes in the content that would add further entries to the resultset thst otherways would be missing in the aliasmap. This is now resolved by queueing up those changes and merging into the resultset.

rombert:
Thanks for the patch @DominikSuess . Based on our offline discussions I now understand the problem and the fix. Reviewing the code I think that we could better reuse the existing lock structure, rather than create new locks and maps.

In case of a 'sync' start the `initializing` lock protects the internal `aliasMap` and does not allow change events to update it. I think the simplest way of solving it is to delay the unlocking of `initializing` until the 'async' loading is complete as well.

By looking at the query methods it does not look like it would block them, so we would not break anything. Would this work for you?

rombert:
I am still getting test errors when running multiple times in a loop:
```
[ERROR] Tests run: 39, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.123 s <<< FAILURE! - in org.apache.sling.resourceresolver.impl.mapping.MapEntriesTest
[ERROR] test_doUpdateAlias(org.apache.sling.resourceresolver.impl.mapping.MapEntriesTest)  Time elapsed: 0.008 s  <<< FAILURE!
java.lang.AssertionError: expected:<1> but was:<0>
	at org.apache.sling.resourceresolver.impl.mapping.MapEntriesTest.test_doUpdateAlias(MapEntriesTest.java:1073)
```

cziegeler:
This has been superseeded by recent work

PR #8, opened 2018-06-14
SLING-7544 Make optimized alias lookup non-blocking

rombert:
I've tried to simplify the async init logic somehow and make sure we use the same 'lock' to guard access everywhere. This is a follow of of #7 , trying a different approach.

I still get a test failure though but I'm not sure if I read the test correctly

```[ERROR] Tests run: 39, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.251 s <<< FAILURE! - in org.apache.sling.resourceresolver.impl.mapping.MapEntriesTest
[ERROR] test_doAddAliasWhileInitializing(org.apache.sling.resourceresolver.impl.mapping.MapEntriesTest)  Time elapsed: 0.117 s  <<< FAILURE!
java.lang.AssertionError: expected null, but was:<{alias=child}>
	at org.apache.sling.resourceresolver.impl.mapping.MapEntriesTest.test_doAddAliasWhileInitializing(MapEntriesTest.java:1214)
```

@DominikSuess  - can you clarify what the the above failing test validates? You added it in 04369d24b831b9e4c7881606b82d67d37fa829a7 but I am not 100% sure why you test this way.

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -45,17 +45,16 @@ import java.util.TimerTask;
 import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.locks.ReentrantLock;
 
 import javax.servlet.http.HttpServletResponse;
 
 import org.apache.sling.api.SlingConstants;
 import org.apache.sling.api.SlingException;
 import org.apache.sling.api.resource.LoginException;
-import org.apache.sling.api.resource.QuerySyntaxException;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ResourceResolver;
 import org.apache.sling.api.resource.ResourceUtil;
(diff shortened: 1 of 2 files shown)


DominikSuess:
@rombert - the test adds an enty while initialize is working and validates that the entry added after initialize has been triggered can be found as soon as isAliasMapInitialized returns true.

rombert:
@DominikSuess - I am curious to understand why the `assertNull` is there. The flow is basically the following:

- initialize map entries (`mapEntries.doInit()`)
- assert that the entries are not initialized yet
- add a map entry (`addResource.invoke(mapEntries`...)
- validate that the map entries contain no alias map for the parent resource of the one just added before

I'm curious why you added the null check (last step). Does this break anything if we get mappings returned this early? Note that the checks are racy - the test seems to assume that the `aliasMap` is not initialized but this may not be true due to the async nature.

Obviously this can also be a bug in my refactoring, so I'm eager to understand the reason behind this check.

rombert:
@DominikSuess  - now I think I know what's going on.

I changed the behaviour so that the `initLock` is released when the `aliasMap` is populated in an async manner, not when the `doInit` method exits.

The `addResource` method (as intended) waits for the `initLock` to be available, and therefore always waits until the initialisation is complete, which means that the alias maps are now always populated in the test method you added.

So my questions are:

1. Is my change conceptually correct and does it still keep the optimisation?
2. Do we need to alter the test you added in the last commit, maybe removing the `assertNull` call, or can we remove it altogether?

DominikSuess:
@rombert - I just wonder how your lock scales as IIUC this will lock a thread for each resource being added. In case a big batch change comes in this might be rather ugly. That was the reason for me to collect intermediate changes in a concurrent list and only lock for the add and merge time of this ‚backlog‘ instead of holding the threads open

rombert:
@DominikSuess - yes, it will wait to add intermediate changes until initialisation is complete. IIUC this should only happen at startup or when the component is restarted. In your experience, does that happen in practice?

DominikSuess:
@rombert - It certainly is an edge case but I have seen import jobs being picked up during startup an automatically creating resources with alias (automated product import). It sure is an edge case but the implementation as is would cause a pretty fatal lock situation that I wanted to get rid off (better then before but would kick in later). 

Keep in mind that the case triggering the optimization was about a multi hour traversal blocking the startup. Waiting for the index should not take that long but still I would want to take the whole preparation out of the the occasion and only have a short lock to activate this map

cziegeler:
This has been superseeded by recent work

PR #12, opened 2019-04-01
SLING-8335 :StringIndexOutOfBoundsException from resolver.map with no…

ashokpanghal:
…n absolute path

check for absolute path at the beginning of the method

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
@@ -689,6 +689,10 @@ public class ResourceResolverImpl extends SlingAdaptable implements ResourceReso
      */
     public Resource resolveInternal(final String absPath, final Map<String, String> parameters) {
         Resource resource = null;
+        if (absPath != null && !absPath.startsWith("/")) {
+            logger.debug("resolveInternal: absolute path expected {} ",absPath);
+            return resource; // resource is null at this point
+        }
         String curPath = absPath;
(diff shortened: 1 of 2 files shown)


rombert:
A SonarQube report for the changes added _only by this pull request_ was generated. Please review it at https://builds.apache.org/job/Sling/job/sling-org-apache-sling-resourceresolver/job/PR-12/1/artifact/target/sonar/issues-report/issues-report-light.html

rombert:
@ashokpanghal - sorry for waiting for so long with this. I am not sure that this is the best way forward, maybe @cziegeler has an idea?

But at a minimum we should have a test which makes that bug is fixed and we don't get regressions.

karlpauls:
@rombert, I think the patch of @ashokpanghal is fine and IIRC, inline with what @cziegeler recommended when I asked him. The point is that we expect an absolute path per api. If it isn't we currently fail with an Exception (not per api) - hence, returning null is the right thing todo. Sorry, I was just lazy and didn't comment early enough :-)

rombert:
Ack, thanks for the confirmation @karlpauls

ashokpanghal:
@karlpauls  Added a test case where a relative path mapped to the original path itself. let me know if this assertion is good enough to validate the fix.

rombert:
A SonarQube report for the changes added _only by this pull request_ was generated. Please review it at https://builds.apache.org/job/Sling/job/sling-org-apache-sling-resourceresolver/job/PR-12/2/artifact/target/sonar/issues-report/issues-report-light.html

PR #13, opened 2019-04-05
SLING-8341 : find extra slashes and ignore redundant delimiteres

ashokpanghal:
find the difference of the redundant occurrences vs actual tokens of delimiter ( slash '/ ')

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.StringTokenizer;
 
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


karlpauls:
@ashokpanghal are you sure that this works in all cases? It seems like there could be a problem if the extra delim count would occur in a part of the absPath that gets cut of due to a shorter path (not sure that this could ever happen but just to be save) - would it maybe be possible to add some tests?

ashokpanghal:
@karlpauls I tested with multiple slashes in path and it works fine. Have not handled the special case of extra slashes in start and towards end of the path right after extension ( which is unlikely in a real scenario) . Added a test case for future regression check.

rombert:
A SonarQube report for the changes added _only by this pull request_ was generated. Please review it at https://builds.apache.org/job/Sling/job/sling-org-apache-sling-resourceresolver/job/PR-13/1/artifact/target/sonar/issues-report/issues-report-light.html

PR #15, opened 2019-07-30
Feature/sling 7768: String Interpolation for ETC Map

schaefa:
Fix for ticket: https://issues.apache.org/jira/browse/SLING-7768 providing a way to make /etc/map configurable for different environments.


pom.xml
```diff
@@ -63,6 +63,32 @@
                     </execution>
                 </executions>
             </plugin>
+<!--
```
cziegeler: Changes in the build should not be part of this PR - unless they are strictly required

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -107,6 +109,19 @@
     @Reference
     EventAdmin eventAdmin;
 
+    /** Event admin. */
```
cziegeler: Comment above seems to be wrong :) Please make the reference greedy and remove the commented out lines below

src/main/java/org/apache/sling/resourceresolver/impl/mapping/Interpolator.java
```diff
@@ -0,0 +1,154 @@
+/*
```
cziegeler: It's ok to have a copy here for getting the functionality in, but we should rather directly use the class from Apache Felix by embedding just that class (and using the maven shade plugin to move it to a different package). But we can do this in a follow up commit

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -140,8 +142,10 @@
 
     private boolean updateBloomFilterFile = false;
 
+    private StringInterpolationProvider stringInterpolationProvider;
```
cziegeler: This can be marked as final

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+)
+public @interface StringInterpolationProviderConfiguration {
+
+    @AttributeDefinition(
+        name = "Placeholder Values",
+        description = "A list of key / value pairs separated by a equal (=) sign. " +
+            "The key is not permitted to contain a '=' sign and the first occurrence of '=' " +
+            "separates the key from the value. If no '=' is found the entire key / value pair " +
+            "is dropped.")
+    String[] placeHolderKeyValuePairs() default {"phv.default.host.name=localhost"};
```
cziegeler: Do we really need such a default value?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderImpl.java
```diff
@@ -0,0 +1,132 @@
...
+import org.osgi.service.component.annotations.Modified;
+import org.osgi.service.metatype.annotations.Designate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Designate(ocd = StringInterpolationProviderConfiguration.class)
+@Component(name = "org.apache.sling.resourceresolver.impl.mapping.StringInterpolationProviderImpl")
```
cziegeler: name attribute is not needed

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderImpl.java
```diff
@@ -0,0 +1,132 @@
...
+    @Override
+    public String substitute(String text) {
+        logger.trace("Substitute: '{}'", text);
+        Object result = Interpolator.replace(text, (type, name, dir) -> {
+            String v = null;
+            if (TYPE_ENV.equals(type)) {
+                v = getVariableFromEnvironment(name);
+            } else if (TYPE_PROP.equals(type)) {
+                v = getVariableFromProperty(name);
+            } else {
```
cziegeler: we should define a type for this, like "config:" and not let all types fallback to this. The fallback prevents adding new types later on (if required)

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderImpl.java
```diff
@@ -0,0 +1,132 @@
...
+    private static final String TYPE_ENV = "env";
+
+    private static final String TYPE_PROP = "prop";
+
+    private static final String DIRECTIVE_DEFAULT = "default";
+
+    /** Logger. */
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    private Map<String, String> placeholderEntries = new HashMap<>();
```
cziegeler: Can be marked as final

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderImpl.java
```diff
@@ -0,0 +1,132 @@
...
+    private BundleContext context;
+
+    // ---------- SCR Integration ---------------------------------------------
+
+    /**
+     * Activates this component (called by SCR before)
+     */
+    @Activate
```
cziegeler: I suggest to change this to a constructor

src/test/java/org/apache/sling/resourceresolver/impl/MockedResourceResolverImplTest.java
```diff
@@ -428,15 +429,15 @@ private Resource buildResource(String fullpath, Iterable<Resource> children, Res
...
-    /**
-     * extract the name from a path.
-     * @param fullpath
-     * @return
-     */
-    private String getName(String fullpath) {
-        int n = fullpath.lastIndexOf("/");
-        return fullpath.substring(n+1);
-    }
+//    /**
```
cziegeler: Please remove the commented code

src/test/java/org/apache/sling/resourceresolver/impl/MockedResourceResolverImplTest.java
```diff
@@ -399,7 +400,7 @@ private Resource buildResource(String fullpath, Iterable<Resource> children) {
     @SuppressWarnings("unchecked")
     private Resource buildResource(String fullpath, Iterable<Resource> children, ResourceResolver resourceResolver, ResourceProvider<?> provider, String ... properties) {
         Resource resource = Mockito.mock(Resource.class);
-        Mockito.when(resource.getName()).thenReturn(getName(fullpath));
+        Mockito.when(resource.getName()).thenReturn(getResourceName(fullpath));
```
cziegeler: ResourceUtil.getName() ?

src/test/java/org/apache/sling/resourceresolver/impl/mapping/EtcMappingMapEntriesTest.java
```diff
@@ -16,12 +16,40 @@
  */
 package org.apache.sling.resourceresolver.impl.mapping;
```
cziegeler: There are a lot of changes in this test and I can't see whether these are just improvements or tests need to be adapted?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/Interpolator.java
schaefa: The class is not exported from the Bundle (AFAI can tell) and so I could not use it.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderConfiguration.java
schaefa: That was here because I wanted to illustrate it. I am going to remove it.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderImpl.java
schaefa: I did not like the fallback either but I wanted to avoid forcing the user to write 'config:' for each value. That said etc-mapping is not easy by any chance and so I guess we add that.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderImpl.java
schaefa: Thanks for point this out

src/test/java/org/apache/sling/resourceresolver/impl/MockedResourceResolverImplTest.java
schaefa: I merge in some changes from master to make it easier to merge and so this is not a change of mine.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/EtcMappingMapEntriesTest.java
schaefa: As far as I can tell I did not change anything here beside that I might have removed commented out code.

src/test/java/org/apache/sling/resourceresolver/impl/MockedResourceResolverImplTest.java
schaefa: Not my change but will do

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderImpl.java
schaefa: You want to make the activate method part of the constructor? Can OCD being part of the constructor?

pom.xml
schaefa: Done

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
schaefa: Done

src/main/java/org/apache/sling/resourceresolver/impl/mapping/Interpolator.java
schaefa: I reverted everything back to the code from Felix with the exception of added logging statements and constants

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
schaefa: Done

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderConfiguration.java
schaefa: Done

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderImpl.java
schaefa: Done

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderImpl.java
schaefa: Done. This required the tests to be adjusted as now $[config: is required.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderImpl.java
schaefa: Opted to replace the Map in activate (modified)

src/test/java/org/apache/sling/resourceresolver/impl/MockedResourceResolverImplTest.java
schaefa: Done

src/test/java/org/apache/sling/resourceresolver/impl/MockedResourceResolverImplTest.java
schaefa: Kept it as this was not my change

src/test/java/org/apache/sling/resourceresolver/impl/mapping/EtcMappingMapEntriesTest.java
schaefa: There are some additional tests cases to deal with types and defaults but the rest did change little.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/Interpolator.java
cziegeler: We can embed it, exactly because it is not exported - but there is no strict need for it, other than to avoid code duplication - and as mentioned we can do this later on

src/main/java/org/apache/sling/resourceresolver/impl/mapping/StringInterpolationProviderConfiguration.java
cziegeler: I still see it :)

PR #16, opened 2019-12-20
SLING-8946 non deterministic resource observation

Buuhuu:
This change calls the same logic that is applied for ResourceProviderHandlers when a ResourceChangeListener is registered after a ResourceProvider got either activated or deactivated.

### DIFF
--- a/src/test/java/org/apache/sling/resourceresolver/impl/providers/ResourceProviderTrackerTest.java
+++ b/src/test/java/org/apache/sling/resourceresolver/impl/providers/ResourceProviderTrackerTest.java
@@ -19,15 +19,21 @@
 package org.apache.sling.resourceresolver.impl.providers;
 
 import static org.hamcrest.Matchers.arrayWithSize;
+import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasSize;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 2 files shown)


Buuhuu:
@rombert @kwin may either of you have a look at this?

rombert:
I think @cziegeler would be the best to review this.

cziegeler:
Long time since I last looked at this code, but I think the changes are good

rombert:
Thanks @cziegeler . @Buuhuu  - can you please squash all commits into a single one and force-push? I will then merge the PR.

src/test/java/org/apache/sling/resourceresolver/impl/providers/ResourceProviderTrackerTest.java
```diff
@@ -240,6 +248,62 @@ public void providerRemoved(String name, String pid, boolean stateful, boolean u
...
+                    excludeSetsPerPath.add(excludesIt.next().getPath());
+                }
+
+                return super.create(path, excludes);
+            }
+        });
+
+        ResourceProvider rp = mock(ResourceProvider.class);
+        ResourceProviderInfo info;
+        // register on RP on /content/B2C, empty exclude set expected
```
Buuhuu: I will remove those project specifics

Buuhuu:
Great thanks. Can we continue the discussion about back-porting this to 1.5.x in the ticket?

PR #20, opened 2020-08-20
SLING-9676 - Improve Javadoc for MapEntriesHandler

rombert:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntry.java
```diff
@@ -268,7 +267,12 @@ public MapEntry(String url, final int status, final boolean trailingSlash,
...
     }
 
-    // Returns the replacement or null if the value does not match
+    /**
+     * Replaces the specified value according to the rules of this entry
+     * 
+     * @param value the value to replace
+     * @return a replaced value of <code>null</code> if the value does not match
+     */
     public String[] replace(final String value) {
```
anchela, line 276: while adding nullability annotations is not in the scope of the original jira ticket, i am very glad to see you added them for the MapEntriesHandler. imo they also help improve readability of the code (apart from automatic nullability violations being spotted and helping to prevent errors)....

so, maybe adding '@Nullable' for the return value and '@NotNull' with the value param would be a good thing.... but as i said: i was mainly asking for javadoc and this would to just an added benefit. but if it leaves the MapEntry class in a odd state because the annotations are only added in some places, feel free to just ignore this comment.

anchela (approved):
looks good to me. thanks a lot... that really helps.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntry.java
rombert: I've added the nullability annotations, it makes sense. I think adding them to just one method helps in the sense that missing annotations don't trigger any incorrect assumptions.

PR #24, opened 2020-11-03
SLING-9535 - Improve performance of sling:alias

akankshajain18:


ankitaagar (approved):

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -128,6 +129,9 @@
 
     private volatile ResourceResolverFactoryConfig config = DEFAULT_CONFIG;
 
+    /** Alias path whitelist */
+    private AtomicReferenceArray<String> aliasPathAllowList;
```
rombert, line 133: I think it's simpler if you have a `volatile` list here and for thread safety either make them immutable , e.g. `Collections.unmodifiableList(...)` or use a thread-safe list like `CopyOnWriteArrayList`.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -304,6 +312,25 @@ protected void activate(final BundleContext bundleContext, final ResourceResolve
             this.observationPaths[i] = new Path(paths[i]);
         }
 
+        // optimize alias path allow list
+        this.aliasPathAllowList = null;
```
rombert, line 316: I think you should swap/null out the field at the end, otherwise there is a window of opportunity when this field is null during recalculation.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
```diff
@@ -149,6 +149,12 @@
                      " and on the alias update time if the number of aliases is huge (over 10000).")
     boolean resource_resolver_optimize_alias_resolution() default true;
 
+    @AttributeDefinition(name = "Allowed Optimize alias path",
```
rombert, line 152: Nit: I think "Allowed optimized alias paths" communicated the intent better.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
```diff
@@ -149,6 +149,12 @@
...
+         description = "This setting can contain a list of path prefixes, e.g. /libs/, /content/. If " +
+                 "such a list is configured, for alias optimization, only paths from resources starting with this prefix " +
+                 "are considered. If the list is empty, all paths are used.)")
+    String[] resource_resolver_optimize_alias_allowedlist();
```
rombert, line 156: Nit: I think "allowlist" is cleared than "allowedlist"

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapConfigurationProvider.java
```diff
@@ -78,4 +78,6 @@ public int compareTo(VanityPathConfig o2) {
      * If <code>null</code> is returned, all paths are allowed.
      */
     List<VanityPathConfig> getVanityPathConfig();
+
+    List<String> getAliasPath();
```
rombert, line 82: I think getAllowedAliasPaths would be better.Also please document what this returns, if returns is nullable and what that means, etc.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1047,13 +1047,40 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
         return map;
     }
 
-	/**
+    private boolean isValidAliasPath(final String path){
```
rombert, line 1050: Indentation looks inconsistent in this method, please check.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
```diff
@@ -2074,4 +2102,28 @@ public void run() {
             }
         }
     }
+
+    @Test
+    public void testIsValidAliasPath() throws Exception {
```
rombert, line 2107: Have you considered making that method have `package` level access by removing the `private` modifier? It might make the code simpler with no ill side-effects.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
```diff
@@ -2074,4 +2102,28 @@ public void run() {
...
+        // ignore system tree - path should not start with /jcr:system -
+        assertFalse((Boolean)method.invoke(mapEntries, "/jcr:system/node"));
+        //valid alias path
+        assertTrue((Boolean)method.invoke(mapEntries, "/parent"));
+        // notallowedparent is not valid configured alias path
+        assertFalse((Boolean)method.invoke(mapEntries, "/notallowedparent"));
+    }
+
+    @Test
+    public void testNullAliasPath() throws NoSuchMethodException, IllegalAccessException {
```
rombert, line 2120: This method does not fail if an exception is not thrown.  You probably want either `@Test(expected=...)` or to use the `ExpectedException` JUnit rule. I you remove the usage of reflection (see comment on previous method) you can use the first option.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1047,13 +1047,40 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
+            }
+
+          // ignore system tree
+          if (path.startsWith(JCR_SYSTEM_PREFIX)) {
+            log.debug("loadAliases: Ignoring {}", path);
+            return false;
+          }
+
+            // check white list
+            if ( this.factory.getAliasPath() != null && !this.factory.getAliasPath().isEmpty()) {
```
rombert, line 1062: getAliasPath() cannot return null, right?

rombert (changes requested):
Thanks for the submission, looks good overall! I think the direction is correct and useful. I have added a few inline comments where I found opportunities for improvement.

I think it would be good to also list the alias paths in the web console page ( http://localhost:8080/system/console/jcrresolver ).

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
jsedding: The equivalent property for vanity paths is called `resource_resolver_vanitypath_whitelist`. For consistency's sake I think this should be called `resource_resolver_alias_whitelist` or maybe slightly more descriptive `resource_resolver_alias_location_whitelist`.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
jsedding: Or similar to the vanity equivalent "Allowed Alias Locations".

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
jsedding: I strongly agree with @rombert. `AtomicReferenceArray` is not the correct data-structure for how you use it. It allows atomic updates to individual array entries, but not for the entire array. You need an atomic update of the entire array (or collection). Indeed, you reassign this field later, which is not thread-safe because the field is not volatile.

If you want to atomically swap out the entire configuration, you need a volatile field and make a local copy of the field before you work with it. Or you use a thread-safe data-structure, like e.g. `CopyOnWriteArrayList` as Robert suggests.

src/main/java/org/apache/sling/resourceresolver/impl/CommonResourceResolverFactoryImpl.java
```diff
@@ -472,6 +473,22 @@ public boolean hasVanityPathPrecedence() {
...
     }
 
+    @Override
+    public List<String> getAliasPath() {
+        final AtomicReferenceArray<String> includes = this.activator.getOptimizedAliasResolutionAllowList();
+        if (includes == null) {
+            return Collections.emptyList();
+        }
+
+        final List<String> configs = new ArrayList<>();
```
jsedding, line 483: I recommend using a `TreeSet`. It provides sorting and eliminates duplicates. Also, I would avoid manual iteration. Instead if `this.activator.getOptimizedAliasResolutionAllowList()` returns a `String[]` you could write `new TreeSet<>(Arrays.asList(includes))`; or if it returns a `Collection` you could just copy it into a `TreeSet` using the constructor.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
jsedding: +1 - either compute the new data-structure and swap it out with a single assingment, or use a concurrent data-structure that could be updated incrementally. I would prefer the former, as it reflects the atomic nature of a config update better.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapConfigurationProvider.java
jsedding: +1 to `getAllowedAliasPaths()` or `getAllowedAliasLocations()`

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1047,13 +1047,40 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
+                boolean allowed = false;
+                for(final String prefix : this.factory.getAliasPath()) {
```
jsedding, line 1064: Nit: this could be much nicer with streams:

`boolean allowed = this.factory.getAliasPath().stream().anyMatch(path::startsWith)`

(Disclaimer: the above was typed in this comment box, so most likely an error is hiding somewhere)

jsedding (changes requested):

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
akankshajain18: As per my understanding, Null is set in the activate method before populating "aliasPathAllowList" to avoid any overlap. Since we are loading this variable using the configured paths, it can not be set null at the end

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
rombert: I would say that since we started discussing about https://cwiki.apache.org/confluence/display/SLING/Removal+of+problematic+language we should not add any more problematic terms like whitelist.

src/main/java/org/apache/sling/resourceresolver/impl/CommonResourceResolverFactoryImpl.java
akankshajain18: @jsedding 
"this.activator.getOptimizedAliasResolutionAllowList()"  return AtomicReferenceArray, as we require atomicity while reading data from config. We can not make "this.activator.getOptimizedAliasResolutionAllowList()" return String[] Since sonar bot suggestion does not allow to have Non-primitive fields as "volatile" and throw a bug in the PR.

With AtomicReferenceArray, we can not use treeSet as it throw "java.lang.ClassCastException:" we may require some comparable method to do so.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
akankshajain18: Initially, Volite keyword is used but sonar bot suggestion suggested to use **AtomicReferenceArray** as should not have Non-primitive fields as "volatile" and throwing bug in the PR.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
akankshajain18: isValidAliasPath() is private method. That' why reflection is used in this test case.
The expected exception will be "InvocationTargetException", Inside catch assert statement has been written

`catch (InvocationTargetException e){
            assertEquals("Unexpected null path", e.getTargetException().getMessage());
        }`

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
akankshajain18: Removed null check

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
rombert: I am not sure why a `volatile CopyOnWriteArrayList` wouldn't work:

- The `volatile` modifier ensures that that the readers of the field always get the latest value - so changing the reference works
- the `CopyOnWriteArrayList` implementation is thread-safe internally, so changes made by one thread are visible to other threads

Maybe SonarCloud flagged a volatile field holding a non-thread-safe list implementation, like `ArrayList`? That would not solve the second problem listed above.

I think using a List is much more easier to understand than an AtomicReferenceArray, so I'd prefer that if at all possible.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
rombert: Ah, I missed the fact that this was done in the `activate` method, so no one can read it at that point. But why do you need to explicitly set it to null? It already has that value since no one has changed it.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
rombert: See my above comment - I suggested _not_ making this method private to ease testing.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
akankshajain18: I followed the initialization done for vanityPath related variables eg. vanityPathWhiteList
So initialized it with Null

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
akankshajain18: yeah understood, made required changes

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
akankshajain18: I tried to list and volatile, in both cases, sonar cloud was throwing a bug saying "should not have Non-primitive fields as volatile, and suggested to use AtomicReferenceArray.
So I modified the PR and used AomicReferenceArray.

wimsymons: I'd suggest to drop the volatile keyword, just use CopyOnWriteArrayList and initialize it here. Let the rest of the code never set it to another value. Just use add/remove/clear. Thread-safe, readable and no Sonar issues.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -304,6 +312,25 @@ protected void activate(final BundleContext bundleContext, final ResourceResolve
...
+        String[] aliasPathPrefix = config.resource_resolver_optimize_alias_allowlist();
+        if ( aliasPathPrefix != null ) {
+            final List<String> prefixList = new ArrayList<>();
+            for(final String value : aliasPathPrefix) {
+                if ( value.trim().length() > 0 ) {
```
wimsymons, line 321: You used `value.trim()` 3 times, I'd suggest to extract it into a local variable.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1047,13 +1023,39 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
+        }
+
+        // ignore system tree
+        if (path.startsWith(JCR_SYSTEM_PREFIX)){
+            log.debug("loadAliases: Ignoring {}", path);
+            return false;
+        }
+        Set<String> allowedPaths = this.factory.getAllowedAliasPaths();
+        // check allow list
+        if(!allowedPaths.isEmpty()){
```
wimsymons, line 1043: Why the extra check? The next statement will do this for you anyway.

wimsymons:
Hi, I reviewed your PR and made a few comments.

But overall, you don't change the JCR query being executed, causing this to be slow for the initial load.

You request *all* `sling:alias` in the repository (including those from `/jcr:system` and those from paths not on your allowed list) and than filter them out in code (inside the `isValidAliasPath` function).

You should change the JCR query by adding something like ISDESCENDANTNODE for every path listed in the `resource.resolver.optimize.alias.allowedlist` OSGi property. This will filter them out *before* they are loaded.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
jsedding: The sonar message is misleading. A more accurate message would say that a volatile field should only hold an immutable object (which includes primitives). So wrapping a `List` using e.g. `Collections.unmodifiableList()` would also be correct and _should not_ trigger any sonar warnings. Alternatively, using a final field holding a thread-safe collection like `CopyOnWriteArrayList` is also correct (in the sense of thread-safety).

Swapping out an immutable object with another in a volatile field constitutes an atomic update, which is often desirable.

Updating a thread-safe collection is not necessarily atomic, so there _could_ be a time-window when the field could be read and the collections contents are in the middle of an update. This may or may not be a problem depending on the use-case.

Using `Atomic*Array` sounds like bad advice to me, because it usually (my opinion) you will not want to swap out the contents of a particular array index atomically. It allows elements to be updated atomically, which is not what you want here.

You want to swap out one configuration for another, so updating a volatile field with an immutable collection sounds most correct to me.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
jsedding: You're right @akankshajain18. However, not all committed code is free from code-smells. I think there's a fine line between emulating the style of an existing code-base, and copying bad practices without questioning them.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
jsedding: Agreed. We can use unproblematic language and communicate the intent more clearly at the same time. What about `resource_resolver_allowed_alias_locations`?

rombert: That would work for me.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
akankshajain18: if allowed path is empty then all path will be considered and the method should return true
 boolean allowed = allowedPaths.stream().anyMatch(path::startsWith);
This statement will return false if allowedPaths will be empty
That's why an empty check is applied explicitly

akankshajain18:
@wimsymons 
The initial thought was changing the query itself.
But as the number path will increase, we need to append ISDESCENDANTNODE with "OR" keyword.
==========
Sample query:
SELECT [sling:alias] FROM [nt:base] AS page
WHERE (ISDESCENDANTNODE(page ,"/content/") OR ISDESCENDANTNODE(page ,"/apps/")) AND [sling:alias] IS NOT NULL
==========

As we are checking multiple paths, will it not slow down the query performance as the path will increase ?

akankshajain18:
The initial thought was changing the query itself.
But as the number path will increase, we need to append ISDESCENDANTNODE with "OR" keyword.

Sample query:
`SELECT [sling:alias] FROM [nt:base] AS page
WHERE (ISDESCENDANTNODE(page ,"/content/") OR ISDESCENDANTNODE(page ,"/apps/")) AND [sling:alias] IS NOT NULL`

As we are checking multiple paths, will it not slow down the query performance as the path will increase?

wimsymons:
> @wimsymons
> The initial thought was changing the query itself.
> But as the number path will increase, we need to append ISDESCENDANTNODE with "OR" keyword.
> 
> Sample query:
> `SELECT [sling:alias] FROM [nt:base] AS page WHERE (ISDESCENDANTNODE(page ,"/content/") OR ISDESCENDANTNODE(page ,"/apps/")) AND [sling:alias] IS NOT NULL`
> 
> As we are checking multiple paths, will it not slow down the query performance as the path will increase?
> 
> Update:
> As i was not sure about the performance of 'ISDESCENDANTNODE', did not follow that path

Performance for ISDESCENDANTNODE is great, even with multiple OR-s. The (Lucene) query engine will take care of that.
You can specify up to a 1000 OR's (this is configurable), but I don't think you will have that many aliasPathAllowList entries. That would be limited to about 1-10 entries. (Most of the time it should only be looking at `/content`).

When you have 0 entries (and target the whole repository), it could even make sense to add an inverted condition not to process `/jcr:system`, like `WHERE NOT ISDESCENDANTNODE(page, "/jcr:system") AND [sling:alias] IS NOT NULL`.

Don't know if the ISDESCENDANTNODE will work when the paths have trailing `/` though. I'm sure `ISDESCENDANTNODE(page ,"/content")` will work, but not sure about `ISDESCENDANTNODE(page ,"/content/")`.

akankshajain18:
> > @wimsymons
> > The initial thought was changing the query itself.
> > But as the number path will increase, we need to append ISDESCENDANTNODE with "OR" keyword.
> > Sample query:
> > `SELECT [sling:alias] FROM [nt:base] AS page WHERE (ISDESCENDANTNODE(page ,"/content/") OR ISDESCENDANTNODE(page ,"/apps/")) AND [sling:alias] IS NOT NULL`
> > As we are checking multiple paths, will it not slow down the query performance as the path will increase?
> > Update:
> > As i was not sure about the performance of 'ISDESCENDANTNODE', did not follow that path
> 
> Performance for ISDESCENDANTNODE is great, even with multiple OR-s. The (Lucene) query engine will take care of that.
> You can specify up to a 1000 OR's (this is configurable), but I don't think you will have that many aliasPathAllowList entries. That would be limited to about 1-10 entries. (Most of the time it should only be looking at `/content`).
> 
> When you have 0 entries (and target the whole repository), it could even make sense to add an inverted condition not to process `/jcr:system`, like `WHERE NOT ISDESCENDANTNODE(page, "/jcr:system") AND [sling:alias] IS NOT NULL`.
> 
> Don't know if the ISDESCENDANTNODE will work when the paths have trailing `/` though. I'm sure `ISDESCENDANTNODE(page ,"/content")` will work, but not sure about `ISDESCENDANTNODE(page ,"/content/")`.

@wimsymons 
I already validated the above query which has trailing '/' .it was working fine.

akankshajain18:
Hi All

Thank you for all the insightful reviews.
I have updated the PR.
Important changes: 
1. use of copyOnWriteArrayList
2. updated sling query.

Kindly review it and share your feedback, if any

wimsymons (changes requested):

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -111,6 +88,8 @@
 
     static final String ALIAS_QUERY_DEFAULT = "SELECT sling:alias FROM nt:base WHERE sling:alias IS NOT NULL";
 
+    static final String ALIAS_BASE_QUERY_DEFAULT = "SELECT sling:alias FROM nt:base As page";
```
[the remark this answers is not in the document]
akankshajain18: To maintain consistency as all keywords are in uppercase

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapConfigurationProvider.java
```diff
@@ -85,5 +84,5 @@ public int compareTo(VanityPathConfig o2) {
      * If empty set is returned, all paths are allowed.
      * @return
      */
-    Set<String> getAllowedAliasPaths();
+    CopyOnWriteArrayList<String> getAllowedAliasPaths();
```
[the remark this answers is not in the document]
akankshajain18: class CopyOnWriteArrayList<E>
    implements List<E>, RandomAccess, Cloneable, java.io.Serializable
When i used List<String>, Sonar cloud was throwing the  bug

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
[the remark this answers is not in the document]
jsedding: I believe @wimsymons was suggesting to write `AS` in all-caps. You have (inconsistently) capitalized the `A` but not the `S`.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -313,21 +306,21 @@ protected void activate(final BundleContext bundleContext, final ResourceResolve
...
                     } else {
-                        prefixList.add(value.trim() + "/");
+                        prefixSet.add(value + "/");
                     }
                 }
             }
-            if ( !prefixList.isEmpty()) {
-                this.aliasPathAllowList = new AtomicReferenceArray<>( prefixList.toArray(new String[prefixList.size()]));
+            if ( !prefixSet.isEmpty()) {
+                this.aliasPathAllowList = new CopyOnWriteArrayList<>(prefixSet);
```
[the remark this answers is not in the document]
akankshajain18: So basically while declaring the CopyOnWriteArrayList, initialization is done and in later parts, add /clear method should be used.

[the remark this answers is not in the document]
jsedding: As I wrote earlier:

> You want to swap out one configuration for another, so updating a volatile field with an immutable collection sounds most correct to me.

Let the field `this.aliasPathAllowedList = Collections.unmodifiableSet(prefixSet)`. You are creating a snapshot (cache) of the configuration values at the time of activation. The contents of the set should never be modified thereafter. On a new configuration update, a new immutable set should be created and atomically assigned (i.e. volatile).

I am not even 100% sure that the `volatile` keyword is required. It probably is, because we have a method annotated with `@Modified`, implying that configuration updates can affect the existing service and thus may happen concurrently.

And yes, @wymsymons is right, only the interface type (in this case `Set` or `Collection`) should be used for fields and return types. That makes reading and refactoring the code easier.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -128,6 +122,9 @@
 
     private volatile ResourceResolverFactoryConfig config = DEFAULT_CONFIG;
 
+    /** Alias path whitelist */
+    private CopyOnWriteArrayList<String> aliasPathAllowList;
```
wimsymons, line 126: @akankshajain18 change this to: 
```
private final List<String> aliasPathAllowList = new CopyOnWriteArrayList<>();
```
to fix the whole discussion about this.

src/main/java/org/apache/sling/resourceresolver/impl/CommonResourceResolverFactoryImpl.java
```diff
@@ -472,6 +466,19 @@ public boolean hasVanityPathPrecedence() {
         return configs;
     }
 
+    @Override
+
+    public CopyOnWriteArrayList<String> getAllowedAliasPaths() {
```
wimsymons, line 471: This should be: ```
public List<String> getAllowedAliasPaths() {
   return Collections.unmodifiableSet(this.aliasPathAllowList);
}
```

File: src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java [hunk 19]
```diff
@@ -304,6 +305,22 @@ protected void activate(final BundleContext bundleContext, final ResourceResolve
             this.observationPaths[i] = new Path(paths[i]);
         }
 
+        // optimize alias path allow list
+        this.aliasPathAllowList.clear();
```
wimsymons, line 309: Hi @akankshajain18, sorry if I might have pointed you in the wrong direction, but now changing the aliasPathAllowList is no longer an atomic operation as @jsedding suggested.

There must be some way to this correctly with the volatile keyword without triggering Sonar errors.

src/main/java/org/apache/sling/resourceresolver/impl/CommonResourceResolverFactoryImpl.java
akankshajain18: @wimsymons 
the idea to return an immutable list is every time the unmodifiable list will be returned and not affected by configuration changes.  After getting all allowed list we need to sort the paths in (` public List<String> getAllowedAliasPaths() ` ), this way we are creating a modifiable list once again
`  List<String> sortedList = immutablelist.stream().sorted().collect(Collectors.toList());`

On other hand sorting paths in the activate method will not be advisable as the activate method should do the minimum required work.

Please suggest

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
akankshajain18: Hi @wimsymons
If I use a volatile keyword with any collection, a sonar bug is triggered.
( Non-primitive fields as volatile is not allowed.) 
That's why I used 'AtomicReferenceArray' as suggested by the sonar cloud.
Later @jsedding and everyone in the review suggested using copyonwriteArraylist.

Please let me know if there is an issue with copyonwriteArraylist.

src/main/java/org/apache/sling/resourceresolver/impl/CommonResourceResolverFactoryImpl.java
jsedding: I suggest that sorting a list with 3 +/- 2 entries is acceptable in the activate method. This "performance hit" is fully compensated by not needing to sort the list repeatedly later on. Come on!

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
jsedding: Tough. Sonar is wrong.

akankshajain18: please check my draft build where I have used volatile keyword
https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/34

akankshajain18:
Hi All

As I have mentioned earlier, using a _virtual keyword with any collection_ causing a sonar bug.
I have created a draft build for reference [https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/34]

To avoid this bug, CopyOnWriteArrayList is used as suggested by everyone in the review.
Please let me know if I have missed something,.
Is there a way out with virtual keyword?

wimsymons:
Don't know if this code goes through the same set of Sonar rules, but I found some look-a-like case at https://github.com/Adobe-Consulting-Services/acs-aem-commons/blob/master/bundle/src/main/java/com/adobe/acs/commons/oak/impl/EnsureOakIndex.java. Here ignoreProperties is also a CopyOnWriteArrayList, which is initialized/atomically assigned in the @Activate method.

jsedding:
@akankshajain18 I do not doubt for a second that sonar flags a _volatile_ field of a type other than a primitive or an enum as a bug. Primitive types and enums have in common that they are _immutable_, and that is what's important here.

Even the Sonar rule's explanation talks about mutability:

`For mutable objects, the volatile should be removed, and some other method should be used to ensure thread-safety [...]`

Now, to add to the confusion, we would like our configuration update not only to be visible by multiple threads (all that the volatile keyword ensures is immediate visibility of the field's update to other threads), but we would also like the update to be _atomic_ (atomic, in case that's not clear, means that there are no intermediate (and possibly invalid) states visible during the update.

Fortunately our configuration object (in this case a set of paths) can be _immutable_, if we wrap the set with `Collections.unmodifiableSet(Set)`. So we can create this immutable set and assign it _atomically_ to the _volatile_ field. This is perfectly good and correct Java.

The fact that Sonar flags this as a bug is a bug in Sonar (or if you prefer a limitation). Java as a language does not have the concept of immutability built in like some other languages do. This makes it very hard for a static analysis tool like Sonar to determine when/if an object is or is not immutable. Owing to the complexity of this problem, it seems that Sonar chose to implement a very simple check instead of a correct one.

Using a `CopyOnWriteArrayList` is IMHO not the correct solution for this problem. As I explained above, we are looking for `immutability` (because that's the characteristic of each state of configuration). However, `CopyOnWriteArrayList` is mutable. Furthermore, we are looking for _atomic_ updates. However, (assuming a _final_ field `CopyOnWriteArrayList`), updates of the `CopyOnWriteArrayList` are not _atomic_, because we would need to clear the previous state (or remove each element individually) and add the new elements.

This is why I continue to advocate a solution which holds an immutable set in a volatile field. Sonar provides a way to ignore errors on a single line. I argue that this is a case where ignoring Sonar for a single line is warranted.

wimsymons:
I agree with Julian here. Great explanation btw :-)

jsedding:
> I agree with Julian here. Great explanation btw :-)

Thanks Wim :)

akankshajain18:
Hi All
I have updated PR as per the review comments.
Please let me know if it is good to go.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -123,7 +123,8 @@
     private volatile ResourceResolverFactoryConfig config = DEFAULT_CONFIG;
 
     /** Alias path whitelist */
-    private final List<String> aliasPathAllowList = new CopyOnWriteArrayList<>();
+    @SuppressWarnings("java:S3077")
+    private volatile Set<String> aliasPathAllowList = Collections.unmodifiableSet(Collections.emptySet());
```
jsedding, line 127: `Collections.emptySet()` is already immutable. No need to wrap it with `Collections.unmodifiableSet()`.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -123,7 +123,8 @@
     private volatile ResourceResolverFactoryConfig config = DEFAULT_CONFIG;
 
     /** Alias path whitelist */
```
jsedding, line 125: Comment still contains the term "whitelist". I would argue the comment is not needed, because the field's name describes its purpose sufficiently.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -306,27 +307,29 @@ protected void activate(final BundleContext bundleContext, final ResourceResolve
...
+            final Set<String> prefixSet = new TreeSet<>();
             for(final String prefix : aliasPathPrefix) {
                 String value = prefix.trim();
-                    if (!value.isEmpty()&&value.startsWith("/")) { // absolute path should be given
-                        if (value.endsWith("/")) {
-                            this.aliasPathAllowList.add(value);
-                        } else {
-                            this.aliasPathAllowList.add(value + "/");
-                        }
+                if (!value.isEmpty()) {
```
jsedding, line 315: What happened to the check for absolute path (starts with "/")? Is it not needed anymore? Why?

If you do re-introduce the check, I would log a warning if it is not an absolute path (but keep skipping empty values silently).

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -306,27 +307,29 @@ protected void activate(final BundleContext bundleContext, final ResourceResolve
...
             }
-            Collections.sort(this.aliasPathAllowList);
         }
 
         // vanity path white list
         this.vanityPathWhiteList = null;
         String[] vanityPathPrefixes = config.resource_resolver_vanitypath_whitelist();
         if ( vanityPathPrefixes != null ) {
-            final List<String> prefixList = new ArrayList<>();
+             List<String> prefixList = new ArrayList<>();
```
jsedding, line 332: I like `final` variables. It tells me I don't have to scan the code for unexpected re-assignments. Also, this line is outside the scope of this PR if I understand correctly.

src/test/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryTest.java
```diff
@@ -108,7 +109,7 @@
     @Test public void testGetAllowedAliasPaths() throws NoSuchMethodException {
         assertTrue(this.commonFactory.getAllowedAliasPaths().isEmpty());
         String[] allowPaths = {"/parent", "/parent0"};
-        setInaccessibleField("aliasPathAllowList", activator, Collections.unmodifiableList(Arrays.asList(allowPaths)));
+        setInaccessibleField("aliasPathAllowList", activator, Collections.unmodifiableSet(new HashSet<>(Arrays.asList(allowPaths))));
```
jsedding, line 112: I would use `TreeSet` again instead of `HashSet`, even in the test. Just in case some logic (now or in the future) relies on the allowed alias locations to be sorted.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
```diff
@@ -116,9 +116,9 @@ public void setup() throws Exception {
                 Collections.<Resource> emptySet().iterator());
         //when(resourceResolverFactory.getAliasPath()).thenReturn(Arrays.asList("/child"));
 
-        CopyOnWriteArrayList<String> aliasPath = new CopyOnWriteArrayList<>();
+        Set<String> aliasPath = new TreeSet<>();
         aliasPath.add("/parent");
-        for(int i = 1;i<testSize;i++){
+              for(int i = 1;i<testSize;i++){
```
jsedding, line 121: Indentation looks wrong. Also the Sling code-base normally uses white-space after the semicolons and around the comparator.

jsedding (commented):
@akankshajain18 The latest changes look pretty good! I think all the concurrency and atomicity aspects are correct now.

I added some comments with minor nits and suggestions - none are blockers IMHO.

And I noticed that there is some inconsistency in the terminology. We have `resource_resolver_allowed_alias_locations`, `getAllowedAliasPaths`, `getOptimizedAliasResolutionAllowList`, `aliasPathAllowList` and maybe more. I guess this is due to the many rounds of refactoring. It would be nice to go over the naming again and to make it consistent. I would opt for variants of "allowedAliasLocations", which aligns everything with the naming of the configuration value's name. In my experience this makes it easier for another developer (or yourself) to read and understand the code in the future.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
akankshajain18: Check is required, Will update PR along with warning message

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
akankshajain18: yes, it is not in the scope of PR.
updated changes

akankshajain18:
> @akankshajain18 The latest changes look pretty good! Thanks for hanging in there! I think all the concurrency and atomicity aspects are correct now.
> 
> I added some comments with minor nits and suggestions - none are blockers IMHO.
> 
> And I noticed that there is some inconsistency in the terminology. We have `resource_resolver_allowed_alias_locations`, `getAllowedAliasPaths`, `getOptimizedAliasResolutionAllowList`, `aliasPathAllowList` and maybe more. I guess this is due to the many rounds of refactoring. It would be nice to go over the naming again and to make it consistent. I would opt for variants of "allowedAliasLocations", which aligns everything with the naming of the configuration value's name. In my experience this makes it easier for another developer (or yourself) to read and understand the code in the future.



> @akankshajain18 The latest changes look pretty good! Thanks for hanging in there! I think all the concurrency and atomicity aspects are correct now.
> 
> I added some comments with minor nits and suggestions - none are blockers IMHO.
> 
> And I noticed that there is some inconsistency in the terminology. We have `resource_resolver_allowed_alias_locations`, `getAllowedAliasPaths`, `getOptimizedAliasResolutionAllowList`, `aliasPathAllowList` and maybe more. I guess this is due to the many rounds of refactoring. It would be nice to go over the naming again and to make it consistent. I would opt for variants of "allowedAliasLocations", which aligns everything with the naming of the configuration value's name. In my experience this makes it easier for another developer (or yourself) to read and understand the code in the future.

All suggested changes are included in the update PR.

jsedding (approved):
Thanks @akankshajain18 for implementing the changes I requested. From my side you have a green light.

wimsymons (approved):

akankshajain18:
> I think it would be good to also list the alias paths in the web console page ( http://localhost:8080/system/console/jcrresolver ).

I don't have much clarity regarding your request for mentioning alias paths on the web console page. Can you please share some details about it?

akankshajain18:
@rombert, PR got a green signal from @jsedding and @wimsymons.
Please have a look and let me know if it is good to go

rombert:
> > I think it would be good to also list the alias paths in the web console page ( http://localhost:8080/system/console/jcrresolver ).
> 
> @rombert
> I don't have much clarity regarding your request for mentioning alias paths on the web console page. Can you please share some details about it?

The web console page at lists various configuration settings. It would be good to also include the new alias paths configuration

image

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1038,22 +1016,69 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
+        logger.debug("Query to fetch alias [{}] ", aliasQuery);
+
+        return aliasQuery;
+    }
+
+    /**
+     *
+     *  validate alias path based on configuration provided
+     */
+    protected boolean isValidAliasPath(final String path){
```
rombert, line 1068: This method looks inconsistent to me - looks like it checks for valid alias paths but returns true unless it gets a null argument, in which case it throws an exception. Do we actually still need it?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -100,6 +76,8 @@
 
     private static final int VANITY_BLOOM_FILTER_MAX_ENTRIES = 10000000;
 
+    private static final Logger logger = LoggerFactory.getLogger(MapEntries.class);
```
rombert, line 79: Either make this upper case or make it an instance field (instance field is probably better from a class reload POV ).

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
```diff
@@ -124,6 +114,15 @@ public void setup() throws Exception {
         when(resourceResolverFactory.isMaxCachedVanityPathEntriesStartup()).thenReturn(true);
         when(resourceResolver.findResources(anyString(), eq("sql"))).thenReturn(
                 Collections.<Resource> emptySet().iterator());
+        //when(resourceResolverFactory.getAliasPath()).thenReturn(Arrays.asList("/child"));
+
+        Set<String> aliasPath = new TreeSet<>();
+        aliasPath.add("/parent");
+        for(int i = 1;i<testSize;i++){
```
rombert, line 121: Whitespace is off here compared to the rest of the class.

rombert (changes requested):
Mostly LGTM. The web console fix can come later so we can merge the initial set of changes.

Once done, please squash and force-push the changes in a single commit.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
akankshajain18: IF the alias path is null then it throw IllegalArgumentException otherwise it returns true.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
akankshajain18: Removed static keyword from here

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
rombert: Yes, I saw that. I was questioning if we really need it. If we do, the name is not correct.

akankshajain18: Initially, this method was responsible for removing or discarding invalid paths based on whitelist items.
Later query to identify alias path is changed and that part of logic is removed from the code.
But I still kept it, for future perspective. (validity of path is checked in one single function).
Currently, this method check if the path is null or not as validation

rombert: Right.

What I am saying is that it never returns false, so checking the return value is not useful. It should be something like `assertValidAliasPath` or `checkValidAliasPath`, have a void return type and only throw the exception.

akankshajain18: I removed that method and put the if condition inside loadalias method itself.Please check my updated PR

rombert (approved):
LGTM, thanks for your patience with this. Before merging, can you please squash all commits into a single one and force-push?

I can't squash + merge from the GitHub UI unfortunately.

akankshajain18:
@rombert Thank you for reviewing.
I can't see any option to squash  + force-push the PR.
I don't have write access to master.

rombert:
@akankshajain18 . What I meant is that you should squash all changes locally and then force-push. This will create a single commit for me to rebase + merge.

akankshajain18:
@rombert 
Thank you for explaining, I have squash and force push the changes in a single commit.
Please check if you can merge the changes now.

rombert:
All good now, thanks @akankshajain18 for your contribution and for your patience.

PR #38, opened 2020-12-16
Sling 9535 dummy

akankshajain18:


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -123,7 +123,7 @@
     private volatile ResourceResolverFactoryConfig config = DEFAULT_CONFIG;
 
     /** Alias path whitelist */
-    //NOSONAR
+    @SuppressWarnings("squid:RSPEC-3077")
```
jsedding, line 126: Try using the rule key "java:S3077", i.e. `@SuppressWarnings("java:S3077")`.

akankshajain18: Thank you @jsedding :)
I will convert list to set and merge changes to the original PR.

PR #42, opened 2021-02-25
SLING-10167 Ensure only valid absolute paths (not ending with "/") are used in JCR queries for loading existing aliases on startup

stefanseifert:
see https://issues.apache.org/jira/browse/SLING-10167

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
@@ -26,6 +26,7 @@ import java.util.*;
 
 import org.apache.commons.collections4.BidiMap;
 import org.apache.commons.collections4.bidimap.TreeBidiMap;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.sling.api.resource.ResourceDecorator;
 import org.apache.sling.api.resource.ResourceResolverFactory;
 import org.apache.sling.api.resource.path.Path;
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -313,10 +314,11 @@ protected void activate(final BundleContext bundleContext, final ResourceResolve
                 String value = prefix.trim();
                 if (!value.isEmpty()) {
                     if (value.startsWith("/")) { // absolute path should be given
+                        // path must not end with "/" to be valid absolute path
                         if (value.endsWith("/")) {
-                            prefixSet.add(value);
+                            prefixSet.add(StringUtils.removeEnd(value, "/"));
```
kwin, line 319: `StringUtils.removeEnd` already has a check so you can remove the surrounding if

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1036,8 +1037,9 @@ private String updateAliasQuery(){
         baseQuery.append(" ").append("WHERE");
 
         if(allowedLocations.isEmpty()){
+            String jcrSystemPath = StringUtils.removeEnd(JCR_SYSTEM_PREFIX, "/");
```
kwin, line 1040: The check performed by http://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#removeEnd-java.lang.String-java.lang.String- is not really necessary here, because you know that the constant ends with "/", right? I would just use `jcrSystemPath.substring(0, jcrSystemPath.length() - 1);` in this case.

stefanseifert: i would like to keep it like this as the code is just better readable this way. this code line is executed once on startup of the bundle, not need to over-optimize it here.

rombert (approved):
LGTM, thanks.

PR #43, opened 2021-03-31
SLING-10269 cache the result of isResourceType()

joerghoh:


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1046,37 +1051,57 @@ public String getParentResourceType(final String resourceType) {
...
-                         superType = this.getParentResourceType(superType);
-                         if (superType != null && superTypesChecked.contains(superType)) {
-                             throw new SlingException("Cyclic dependency for resourceSuperType hierarchy detected on resource " + resource.getPath(), null);
-                         }
-                     }
-                 }
+
+             // Check if the result is already available from cache
+             String3Tupel key = new String3Tupel(resource.getResourceType(),resource.getResourceSuperType(), resourceType);
+             if (resourceTypeLookupCache.containsKey(key)) {
```
rombert, line 1057: I would suggest a get and null check. This way we don't query the map twice.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1046,37 +1051,57 @@ public String getParentResourceType(final String resourceType) {
     public boolean isResourceType(final Resource resource, final String resourceType) {
         boolean result = false;
         if ( resource != null && resourceType != null ) {
-             // Check if the resource is of the given type. This method first checks the
```
rombert, line 1049: This inline comment seems to have been lost in the refactoring.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1117,4 +1142,68 @@ public Resource move(final String srcAbsPath, final String destAbsPath) throws P
         }
         return rsrc;
     }
+
+
+
+    // A very simple tupel implementation which can be used as key in any map
+    public class String3Tupel {
```
rombert, line 1149: Nit: tuple, not tupel :-)

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1117,4 +1142,68 @@ public Resource move(final String srcAbsPath, final String destAbsPath) throws P
...
+
+        @Override
+        public int hashCode() {
+            final int prime = 31;
```
rombert, line 1153: I think you can use `java.util.Objects.hashCode` here.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1117,4 +1142,68 @@ public Resource move(final String srcAbsPath, final String destAbsPath) throws P
...
+            int result = 1;
+            result = prime * result + getEnclosingInstance().hashCode();
```
rombert, line 1155: Is this necessary? A tuple3 does not escape the scope of a single resource resolver, right?

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1117,4 +1142,68 @@ public Resource move(final String srcAbsPath, final String destAbsPath) throws P
...
+            result = prime * result + ((s1 == null) ? 0 : s1.hashCode());
+            result = prime * result + ((s2 == null) ? 0 : s2.hashCode());
+            result = prime * result + ((s3 == null) ? 0 : s3.hashCode());
+            return result;
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (this == obj)
```
rombert, line 1164: I think you can use `java.util.Objects.equals` here.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1117,4 +1142,68 @@ public Resource move(final String srcAbsPath, final String destAbsPath) throws P
...
+                return true;
+            if (obj == null)
+                return false;
+            if (getClass() != obj.getClass())
+                return false;
+            String3Tupel other = (String3Tupel) obj;
+            if (!getEnclosingInstance().equals(other.getEnclosingInstance()))
```
rombert, line 1171: Same as above, is this needed?

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1117,4 +1142,68 @@ public Resource move(final String srcAbsPath, final String destAbsPath) throws P
...
+                return false;
+            if (s3 == null) {
+                if (other.s3 != null)
+                    return false;
+            } else if (!s3.equals(other.s3))
+                return false;
+            return true;
+        }
+
+        String s1;
```
rombert, line 1191: Any reason to not make these fields private? Also, fields should go before contructors, constructors before instance methods :-)

src/test/java/org/apache/sling/resourceresolver/impl/ResourceResolverImplTest.java
```diff
@@ -521,6 +527,28 @@ public void testCloseWithStackTraceLogging() throws Exception {
         assertFalse(resolver.isResourceType(r, "h:p"));
     }
 
+    @Test public void testIsResourceTypeCached() throws NoSuchFieldException, SecurityException, IllegalArgumentException, IllegalAccessException {
```
rombert, line 530: Nit: maybe `throws Exception`?

rombert (commented):
I did an initial pass on the implementation, overall LGTM, but - as always :-) - I have some minor tweaks I think would be good to include.

There are two major questions outstaning for me:
- is the thread safety enough? We typically don't share resource resolvers between threads, but I think the current contract is "you might get stale data". Unsynchronised access to a HashMap might lead to deadlocks
- are we aware of any inconsistent cache scenarios? I _think_ that all changes done during the current session will be reflected in the current ResourceResolver but maybe it's worth analysing this as well.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1046,37 +1051,57 @@ public String getParentResourceType(final String resourceType) {
...
-                         superTypesChecked.add(superType);
-                         superType = this.getParentResourceType(superType);
-                         if (superType != null && superTypesChecked.contains(superType)) {
-                             throw new SlingException("Cyclic dependency for resourceSuperType hierarchy detected on resource " + resource.getPath(), null);
-                         }
-                     }
-                 }
+
+             // Check if the result is already available from cache
+             String3Tupel key = new String3Tupel(resource.getResourceType(),resource.getResourceSuperType(), resourceType);
```
kwin, line 1056: Instead of a generic 3 string tuple, a more specific class like `ResourceTypeInformation` with speaking getter method names are way easier to understand.

joerghoh: I renamed the class. We don't need any getters, as it is just used as a key in the map.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
joerghoh: Added it as method description for `isResourceTypeInternal`

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
joerghoh: Done

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
joerghoh: Eclipse created both `hashCode` and `equals`, so I didn't want to bother with it :-)

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
joerghoh: Changed order (not sure what Eclipse did here ...)

joerghoh (commented):
@rombert: you raised an interesting point: Changes which are performed via ResourceResolver. While I drop now the cache on every commit, It's actually quite hard to do if someone is doing a change directly via a enclosed JCR session. I don't think that there is a way for sling to detect that. Even the [JcrValueMap](https://github.com/apache/sling-org-apache-sling-jcr-resource/blob/master/src/main/java/org/apache/sling/jcr/resource/internal/JcrValueMap.java) is caching properties of JCR nodes within a ValueMap, and doesn't care if it's in sync with the JCR (and it's not synchronized either).

Besides that: Thanks @bdelacretaz, @rombert , @kwin  for your feedback, everything should be addressed.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
rombert: Well, the maintenance/review cost will be there, so you can ask Eclipse to do the right thing with recent versions :-)

https://stackoverflow.com/a/51775296/112671

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
rombert: I am still missing the point of this field, seems it's only used in hashcode/equals - is it really necessary?

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -124,6 +129,7 @@ private ResourceResolverImpl(final ResourceResolverImpl resolver, final Map<Stri
         this.context = new ResourceResolverContext(this, factory.getResourceAccessSecurityTracker());
         this.control = createControl(factory.getResourceProviderTracker(), authInfo, resolver.control.isAdmin());
         this.factory.register(this, control);
+        this.resourceTypeLookupCache = new HashMap<>();
```
rombert, line 132: This one is still unsynchronized, but see new comment below.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1046,37 +1053,60 @@ public String getParentResourceType(final String resourceType) {
...
-                         superType = this.getParentResourceType(superType);
-                         if (superType != null && superTypesChecked.contains(superType)) {
-                             throw new SlingException("Cyclic dependency for resourceSuperType hierarchy detected on resource " + resource.getPath(), null);
-                         }
-                     }
-                 }
+
+             // Check if the result is already available from cache
+             ResourceTypeInformation key = new ResourceTypeInformation(resource.getResourceType(),resource.getResourceSuperType(), resourceType);
+             Boolean value = resourceTypeLookupCache.get(key);
```
rombert, line 1059: Thinking about it some more, maybe a `ConcurrentMap` is a better fit? It's more expressive, e.g.

```java
resourceTypeLookupCache.computeIfAbset(key, (key) -> isResourceTypeInternal(resource, resourceType));
```

and also skips the locking of the synchronized map.

If you're happy with the synchronized map, feel free to use that.

joerghoh: If the access is normally happening synchronously, the overhead of locking seems acceptable to me compared to the overhead of a more complex implementation. But unless we have benchmarked it's probably an academic discussion :-)

joerghoh:
@rombert: Thanks for your feedback, I adjusted the implementation for `hashcode` as requested (and I wonder why Eclipse needs to be told explicitly to make it right).
I also switched to a ConcurrentHashMap, and the use `computeIfAbsent` makes the code indeed much more expressive (and shorter).

rombert (approved):
LGTM, thanks!

rombert:
@joerghoh - all good know from my POV. Please make sure to _Squash and merge_ when done to keep a single commit for this atomic change.

PR #46, opened 2021-06-02
SLING-10447 Improve the querys that are used to load vanity paths, by specifying path restrictions

henrykuijpers:


rombert:
Thanks for the PR @henrykuijpers ! I'll not merge it immediately as @kwin has some concerns on whether this is the right thing to do.

If we end up going this way, we will need the branch to be rebased as it has conflicts.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1032,34 +1070,33 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
-        baseQuery.append(" ").append("WHERE");
+        StringBuilder query = new StringBuilder(ALIAS_BASE_QUERY_DEFAULT);
+        query.append(" ").append("WHERE");
 
         if(allowedLocations.isEmpty()){
-            String jcrSystemPath = StringUtils.removeEnd(JCR_SYSTEM_PREFIX, "/");
-            baseQuery.append(" ").append("(").append("NOT ISDESCENDANTNODE(page,")
-                    .append("\"").append(jcrSystemPath).append("\"").append("))");
+            query.append(" ").append("(").append("NOT ISDESCENDANTNODE(page,")
+                    .append("\"").append(JCR_SYSTEM_PATH).append("\"").append("))");
```
jsedding, line 1081: It would be nice to reformat this query to use single quotes instead of escaped double quotes. Then both queries share the same style, and it is more readable.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1255,6 +1279,33 @@ private boolean loadVanityPath(final Resource resource, final Map<String, List<M
         return hasVanityPath;
     }
 
+    private String createVanityPathQuery() {
+        final String query =
+            VANITY_PATH_BASE_QUERY_DEFAULT +
+                " WHERE sling:vanityPath IS NOT NULL" +
+                Stream
+                    .of(true, false)
+                    .map(this::createVanityPathQueryPathRestriction)
```
jsedding, line 1288: After reading the code generating the restrictions several times, I believe it is actually correct. But it is not very readable or intuitive.

Maybe it would become more readable if you separated out all "include" restrictions and all "exclude" restrictions, each into their own collection, and then join them with the appropriate "AND" or "OR" operators?

Pseudo code:
```
includesExcludes = Stream.of(new VanityPathConfig(JCR_SYSTEM_PATH, true), factory.getVanityPathConfig().stream())
    .collect(Collectors.groupingBy(VanityPathConfig::isExclude)

excludes = includesExclude.get(Boolean.TRUE)
includes =  includesExclude.get(Boolean.FALSE)

excludesPart =excludes.stream().map("NOT ISDESCENDANTNODE(page, '%s')").join(" AND ")
includesPart = includes.stream().map("ISDESCENDANTNODE(page, '%s')").join(" OR ")

queryRestrictions = String.format("(%s) AND (%s)", excludesPart, includesPart)
```

Beware, I may have gotten the "OR"s and "AND"s wrong ;)

PS: maybe also move `new VanityPathConfig(JCR_SYSTEM_PATH, true)` into `factory.getVanityPathConfig()`?

jsedding (changes requested):

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
```diff
@@ -112,11 +112,6 @@
                       "the ResourceResolver mapping. The default value is /etc/map.")
     String resource_resolver_map_location() default MapEntries.DEFAULT_MAP_ROOT;
 
-    @AttributeDefinition(name = "Mapping Observation",
```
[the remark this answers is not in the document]
henrykuijpers: The mapping observation configuration setting is used to register the `ResourceChangeListener`, so that the `MapEntries` will be notified of changes that it is interested in. In AEM, it is set to `/`. 

However, given the various configurations that exist, it is easy to compute the right paths to use. (That is also changed in the part where the listener is registered.)

There are also some fallbacks in there to listen to `/` anyway, if that is required.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
henrykuijpers: Good points! I will have another look at this. Also, a good point about moving the `/jcr:system`-exclude to the factory.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
henrykuijpers: Agree, I'll do that. I tried to touch as less as possible, but I very much agree with this.

henrykuijpers: I just changed this to use single quotes.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
henrykuijpers: I just had another go at it, @jsedding , does this already look a bit better?

henrykuijpers:
Just did a commit with some fixes, @jsedding . :) Can you have another look?

src/main/java/org/apache/sling/resourceresolver/impl/CommonResourceResolverFactoryImpl.java
```diff
@@ -516,6 +494,16 @@ public ResourceProviderTracker getResourceProviderTracker() {
         return authenticationInfo;
     }
 
+    @Override
```
cziegeler, line 497: Can we maybe use more inclusive language here like allow/deny list?

henrykuijpers: Updated

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
```diff
@@ -370,26 +381,31 @@ public void test_vanity_path_updates() throws Exception {
...
-        when(resourceResolver.findResources(anyString(), eq("sql"))).thenAnswer(new Answer<Iterator<Resource>>() {
-
-            @Override
-            public Iterator<Resource> answer(InvocationOnMock invocation) throws Throwable {
-                if (invocation.getArguments()[0].toString().contains("sling:vanityPath")) {
-                    return resources.iterator();
-                } else {
-                    return Collections.<Resource> emptySet().iterator();
-                }
+        when(resourceResolver.findResources(anyString(), eq("sql"))).thenAnswer((Answer<Iterator<Resource>>) invocation -> {
```
bdelacretaz, line 389: Although you are checking that the query is what you expect, unless I missed something there's no test that validates the query syntax? I understand this is similar to the existing tests but I think it would be better to run that query against a content repository provided by Sling JCR Mocks and validate the results.
I don't know how much time you have for that - this way of testing probably works for now but IMHO running the query would create a test that's easier to maintain and understand. Maybe factor out the query generation to its own package-local helper class to avoid issues with private methods.

henrykuijpers: As far as I know, the execution of querys isn't implemented anywhere in the testing code that we have a available. IIUC, we need an actual instance to be able to execute querys and thus validating the query syntax and also the proper execution of the query. 

What I've done is indeed making sure that the expected query is being executed. 

I think factoring out query code etc etc might be one mile too far for what we're trying to achieve in this ticket? For us, given our big version storage, it's quite important that those paths will not be considered anymore.

bdelacretaz: If other reviewers agree to merge this with the tests as is I will not object. Maybe just add a comment to the test explaining that actually running the query would be better?

henrykuijpers: Just added that comment anyway. :) 

@bdelacretaz  Do you know of a way to test querys with a real query engine, without booting a Sling instance?

joerghoh: You could use SlingContext and back it for the test with a Oak Repo.

rombert: I would be a bit cautious with pulling in too much into this test module. Bring in Sling Mocks in low-level modules has in my experience lead to various classpath clashes and overriding of depedencies.

I think validating the query is better done at the integration test level. Or maybe, if we want to be sure, we might want to try importing a query parser from Oak and checking that the query is valid. But I think that running the full query is too much at this level.

bdelacretaz: The [jcr/repoinit/ExecutionOrderTest](https://github.com/apache/sling-org-apache-sling-jcr-repoinit/blob/master/src/test/java/org/apache/sling/jcr/repoinit/ExecutionOrderTest.java) is an example using SlingContext. I agree with Robert that such things might be better tested in integration tests, and as mentioned I won't object to committing the code without a query validation test.

rombert: For validating queries we can probably wrap the logic from Oak's QueryEngineImpl#parseQuery ( https://github.com/apache/jackrabbit-oak/blob/fa54797c66e262d07e6e139526855ae8cb800bd1/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java#L144-L230 ), which basically delegates to the `SQL2Parser` and `XPathToSQL2Converter` classes. Those classes look pretty self-contained to me and we should be able to stub any dependencies they have - they don't seem used.

To me, this looks like a good idea for a separate module or bundle, so we can reuse it in other places where it may be desirable to do so.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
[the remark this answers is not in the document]
rombert: I think it may be confusing to remove this setting. Perhaps mark it as deprecated and indicate that it is actually auto-computed?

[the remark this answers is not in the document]
rombert: Might also be worth logging a warning if a non-default setting is found.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
henrykuijpers: Indeed, it looks like `ResourceResolverType.JCR_OAK` is a solution that can actually parse and execute a query. Very cool! I've been using it in the project I'm working on and that's really nice. It's also nice that I was able to get com.day.cq.search working in the unit test, to test the predicate search that we're executing. But that's something outside Sling, of course.

I've created https://issues.apache.org/jira/browse/SLING-10742 to tackle this.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -707,7 +748,7 @@ public void onChange(final List<ResourceChange> changes) {
             log.debug("onChange, type={}, path={}", rc.getType(), path);
 
             // don't care for system area
-            if (path.startsWith(JCR_SYSTEM_PREFIX)) {
+            if (path.startsWith(JCR_SYSTEM_PATH + '/')) {
```
cziegeler, line 751: while this is not a big deal, why did you change this from a constant to an expression that is executed each time?

henrykuijpers: That is an oversight: I think I inlined a few values, I'll change it back to a constant.

reschke:
This PR can be closed, right? (see https://issues.apache.org/jira/browse/SLING-10447)

PR #47, opened 2021-06-17
SLING-10476 Aliases in MapEntries are removed in delete event of jcr:content present in the parent directory

sagarmiglani:


rombert:
LGTM, thanks for your contribution @sagarmiglani

PR #48, opened 2021-06-25
SLING-10483 Incorrect mapping is returned when a resourceResolver.map is called with all its aliases

sagarmiglani:


src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
```diff
@@ -110,6 +110,7 @@ public void prepare() throws LoginException {
         resourceProvider.putResource("/content/virtual/foo"); // matches virtual.host.com.80 mapping entry
         resourceProvider.putResource("/parent", PROP_ALIAS, "alias-parent"); // parent has alias
         resourceProvider.putResource("/parent/child", PROP_ALIAS, "alias-child"); // child has alias
+        resourceProvider.putResource("/parent/child/grandChild", PROP_ALIAS, "alias-grandChild"); // grandChild child has alias
```
rombert, line 113: Does this reproduce only with 3+ path segments? I think we already have a scenario. with two segments

src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
```diff
@@ -150,7 +150,9 @@ public String getMapping(String resourcePath, HttpServletRequest request) {
         if (nonDecoratedResource != null) {
             List<String> aliases = loadAliasesIfApplicable(nonDecoratedResource);
             // avoid duplicating the originally requested path
-            aliases.remove(mappedPath);
+            if ( aliases.contains(mappedPath) ) {
```
rombert, line 153: This only influences the order of the returned mappings, right? I think the fix is a bit counter-intuitive and makes the (already complicated code) harder to understand.

It would be great if you could find another way to adjust the order.

rombert (changes requested):
Thanks for the PR @sagarmiglani ! I think this goes in the right direction but there are a couple of things we should address before merging.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
sagarmiglani: Hi Robert,

**_>> Does this reproduce only with 3+ path segments?_**
No, this can also be reproduced with 2 segments, just to replicate the scenario mentioned in the SLING-10483, I added 3 segements.
But, I believe I should have used only 2 segments. WDYT?

**_>> I think we already have a scenario. with two segments_**
Yes, we have an already existing test with two segments (mapResourceWithNestedAlias), but this bug is reproducible only when we have all aliases segments present in the requested resource path.
And in the already existing test, the requested path does not consists of all aliases segments.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
sagarmiglani: Yes, this only influences the order of returned mappings, my analysis is as follows: (please let me know if you think there is some issue with the same)

Lets say we have the following resource structure
		

- /parent -> with alias "parent-alias"
  - /child -> with alias "child-alias"

For the following resource path requests, the primary mapping returned should be:
| Requested resource path  | Expected mapping |
| ------------- | ------------- |
| /parent/child   | /parent-alias/child-alias |
| /parent/child-alias  | /parent-alias/child-alias |
| /parent-alias/child-alias | /parent-alias/child-alias |
| /parent-alias/child-alias (this is our case of concern) | /parent-alias/child-alias |

When the requested path is **/parent-alias/child-alias** (all segments are aliases), our expectation is that the primary returned mapping should be **/parent-alias/child-alias**.
With the above requested path, In aliases list ([ResourceMapperImpl.java:153](https://github.com/apache/sling-org-apache-sling-resourceresolver/blob/master/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java#L153)), /parent-alias/child-alias was present at 0 index and should have been returned as the primary path.
But [ResourceMapperImpl.java:153](https://github.com/apache/sling-org-apache-sling-resourceresolver/blob/master/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java#L153), we are removing the requested path from the aliases list itself, because we have already added the requested path in the mapping at [ResourceMapperImpl.java:143](https://github.com/apache/sling-org-apache-sling-resourceresolver/blob/master/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java#L143). This was making the order of the returned mappings incorrect. 
So, I thought that we should not be removing the requested resource path from the aliases list, instead if aliases list contains the requested resource path, we should remove it from mapping list.

But, If you think this has made the code harder to understand, I'll try to come up with some other approach, but do you have any suggestions for the same?

PR #49, opened 2021-08-26
SLING-10447 Adjust query so that sling:vanityPath properties are not searched in the JCR version storage

henrykuijpers:


rombert:
Thanks for the contribution @henrykuijpers and for the review @cziegeler

PR #50, opened 2021-10-16
SLING-10844: ResourceMapper.getMapping() returns null for empty path

Gabiesfat:
This PR has the code changes for issue SLING-10844. Apparently earlier implementation of API getMapping() in ResourceMapperImpl.java was returning null in case the path was empty which was breaking the contract mentioned in corresponding interface so to fix code changes been made to return empty string (but not null).  
The test [0] would be fixed in further commit of this PR.
[0]:  https://github.com/apache/sling-org-apache-sling-resourceresolver/blob/master/src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java#L173

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
@@ -154,7 +154,7 @@ public class ResourceMapperImpl implements ResourceMapper {
         }
 
         // 5. add the requested path itself, if not already populated
-        if (!mappings.contains(mappedPath))
+        if ( !mappedPath.isEmpty() && !mappings.contains(mappedPath))
             mappings.add(0, mappedPath);
         
         // 6. add vanity paths
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
@@ -72,7 +72,7 @@ public class ResourceMapperImpl implements ResourceMapper {
         
         Collection<String> mappings = getAllMappings(resourcePath, request);
         if ( mappings.isEmpty() )
-            return null;
+            return "";
         
         return mappings.iterator().next();
     }
@@ -154,7 +154,7 @@ public class ResourceMapperImpl implements ResourceMapper {
         }
 
         // 5. add the requested path itself, if not already populated
-        if ( !mappedPath.isEmpty() && !mappings.contains(mappedPath))
+        if (!mappings.contains(mappedPath))
             mappings.add(0, mappedPath);
         
         // 6. add vanity paths
--- a/src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
+++ b/src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
@@ -105,7 +105,7 @@ public class ResourceMapperImplTest {
 
         InMemoryResourceProvider resourceProvider = new InMemoryResourceProvider();
 
-        if(!testName.getMethodName().contains("mapEmptyPathWithNonExistingResource")) {
+        if(!testName.getMethodName().contains("mapEmptyPathWithUnreadableRoot")) {
             resourceProvider.putResource("/"); // root
         }
         resourceProvider.putResource("/here"); // regular page
@@ -165,11 +165,11 @@ public class ResourceMapperImplTest {
     public void mapNonExistingEmptyPath() throws LoginException {
 
         ExpectedMappings.nonExistingResource("")
-                .singleMapping("/")
-                .singleMappingWithRequest("/app/")
-                .allMappings("/")
-                .allMappingsWithRequest("/app/")
-                .verify(resolver, req);
+            .singleMapping("/")
+            .singleMappingWithRequest("/app/")
+            .allMappings("/")
+            .allMappingsWithRequest("/app/")
+            .verify(resolver, req);
     }
 
     /**
@@ -181,11 +181,11 @@ public class ResourceMapperImplTest {
     public void mapNonExistingPath() throws LoginException {
 
         ExpectedMappings.nonExistingResource("/not-here")
-                .singleMapping("/not-here")
-                .singleMappingWithRequest("/app/not-here")
-                .allMappings("/not-here")
-                .allMappingsWithRequest("/app/not-here")
-                .verify(resolver, req);
+            .singleMapping("/not-here")
+            .singleMappingWithRequest("/app/not-here")
+            .allMappings("/not-here")
+            .allMappingsWithRequest("/app/not-here")
+            .verify(resolver, req);
     }
 
     /**
@@ -197,11 +197,11 @@ public class ResourceMapperImplTest {
     public void mapExistingPath() throws LoginException {
 
         ExpectedMappings.existingResource("/here")
-                .singleMapping("/here")
-                .singleMappingWithRequest("/app/here")
-                .allMappings("/here")
-                .allMappingsWithRequest("/app/here")
-                .verify(resolver, req);
+            .singleMapping("/here")
+            .singleMappingWithRequest("/app/here")
+            .allMappings("/here")
+            .allMappingsWithRequest("/app/here")
+            .verify(resolver, req);
     }
 
     /**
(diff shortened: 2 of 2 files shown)


rombert:
Thanks for the PR @Gabiesfat and apologies for looking so late into this. At this point I am not sure what we are trying to fix. I think it would be best if you could demonstrate it with a unit test that fails without the change and passed with it.

I tried to do so by adding a new test case to `ResourceMapperImplTest.java`:

```java
    @Test
    public void mapEmptyPath() {
        ExpectedMappings.nonExistingResource("")
            .singleMapping("/")
            .singleMappingWithRequest("/app/")
            .allMappings("/")
            .allMappingsWithRequest("/app/")
            .verify(resolver, req);
    }
```

The test passes currently so I am probably misunderstanding what this PR is trying to fix.

Gabiesfat:
> Thanks for the PR @Gabiesfat and apologies for looking so late into this. At this point I am not sure what we are trying to fix. I think it would be best if you could demonstrate it with a unit test that fails without the change and passed with it.
> 
> I tried to do so by adding a new test case to `ResourceMapperImplTest.java`:
> 
> ```java
>     @Test
>     public void mapEmptyPath() {
>         ExpectedMappings.nonExistingResource("")
>             .singleMapping("/")
>             .singleMappingWithRequest("/app/")
>             .allMappings("/")
>             .allMappingsWithRequest("/app/")
>             .verify(resolver, req);
>     }
> ```
> 
> The test passes currently so I am probably misunderstanding what this PR is trying to fix.
I tried to make sure that method _getMapping_  returns empty mapping instead of null incase of _resourcePath_ provided as empty string. I have limited understanding of unit tests here as this is the first time I am seeing this code, if you could  help me in writing a test case where empty string returns after providing empty _resourcePath_ then it would be great.

rombert:
@Gabiesfat - that was my point, this test case shows that for an empty path you don't get a null result. It's possible that the unit tests are incorrect somehow. Maybe it's best that you debug your scenario with the empty path, see what code paths are triggered, and reproduce them with a test case.

If you have problems with the test case I can definitely help, but we need to understand the failure scenario first.

rombert (changes requested):
Not to be merged until the exact problem is clarified.

Gabiesfat:
> @Gabiesfat - that was my point, this test case shows that for an empty path you don't get a null result. It's possible that the unit tests are incorrect somehow. Maybe it's best that you debug your scenario with the empty path, see what code paths are triggered, and reproduce them with a test case.
> 
> If you have problems with the test case I can definitely help, but we need to understand the failure scenario first.

Apparently the issue was at [0] the resolver could also resolve to null resource and the same scenario has not been replicated in the test you mentioned. I have added a test which replicates the same scenario and the same fails without my  fix. Please have a look at it let me know your thoughts.
[0]: https://github.com/apache/sling-org-apache-sling-resourceresolver/blob/master/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java#L145

src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
```diff
@@ -355,6 +358,31 @@ public void mapNestedAliasTarget() {
...
                 .verify(resolver, req);
     }
+    
+    /**
+     * Validates that mappings for an empty path
+     *
+     * @throws LoginException
+     */
+    @Test
+    public void mapEmptyPathWithNonExistingResource() {
```
rombert, line 368:
```suggestion
    public void mapEmptyPathWithUnreadableRoot() {
````

File: src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java [hunk 2]
```diff
@@ -355,6 +358,31 @@ public void mapNestedAliasTarget() {
...
+        MapEntriesHandler mapEntriesHandler = mock(MapEntriesHandler.class);
+        when(mapEntriesHandler.getVanityPathMappings()).thenReturn(Collections.emptyMap());
+
+        ResourceResolverImpl resolver = mock(ResourceResolverImpl.class);
+        when(resolver.resolveInternal(any(), any())).thenReturn(null);
```
rombert, line 373: Can we make this something return `null` only for the root path and delegate to the real resolver? I think that's a more likely scenario.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
```diff
@@ -355,6 +358,31 @@ public void mapNestedAliasTarget() {
...
+        when(resolver.getResource("")).thenReturn(this.resolver.getResource(""));
+        when(resolver.adaptTo(ResourceMapper.class)).thenReturn(new ResourceMapperImpl(resolver, mock(ResourceDecoratorTracker.class),
+                mapEntriesHandler, mock(Object.class)));
+
+        ExpectedMappings.nonExistingResource("")
+                .singleMapping("")
+                .singleMappingWithRequest("")
+                .allMappings("")
+                .allMappingsWithRequest("")
+                .testingEmptyPathWithNonExistingResource(true)
```
rombert, line 383: Why is this needed?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
```diff
@@ -72,7 +72,7 @@ public String getMapping(String resourcePath, HttpServletRequest request) {
         
         Collection<String> mappings = getAllMappings(resourcePath, request);
         if ( mappings.isEmpty() )
-            return null;
+            return "";
```
rombert, line 75: I would prefer that we handle this differently, in the `getAllMappings` method. The javadoc of that mapping promises that no empty results will be returned.

https://github.com/apache/sling-org-apache-sling-api/blob/org.apache.sling.api-2.24.0/src/main/java/org/apache/sling/api/resource/mapping/ResourceMapper.java#L109

rombert (changes requested):

src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
humbledev123: Finding an appropriate mappings relies on having non-empty mappedPath or non-null nonDecoratedResource but in case if mappedPath is empty and nonDecoratedResource is null (it could be possible based on doc [0]) then there could be no possible mapping and the result would be empty. Besides I cold not think of any possible mapping which ought to be returned in such case, any thoughts ?

[0]: https://github.com/apache/sling-org-apache-sling-resourceresolver/blob/deb25ba47cf563e5bc52978eef36e80d735a067c/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java#L718

src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
humbledev123: This is for the case where resource is non-existing with empty path.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
humbledev123: Apparently root path is not empty this was done to replicate the scenario which was observed in bug where path is empty and then resolveInternal ought to return null.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
humbledev123: Done

rombert: @humbledev123 where is this done? I don't see any new commits added to this PR.

rombert:
@humbledev123 - thanks for your comments. I still think we need a clear test case or a reliable way of reproducing the problem in the Sling Starter. Then we can see where the fix should land.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
```diff
@@ -100,7 +104,10 @@ public void prepare() throws LoginException {
         ctx.registerInjectActivateService(new StringInterpolationProviderImpl());
 
         InMemoryResourceProvider resourceProvider = new InMemoryResourceProvider();
-        resourceProvider.putResource("/"); // root
+
+        if(!testName.getMethodName().contains("mapEmptyPathWithUnreadableRoot")) {
```
Gabiesfat, line 108: This was done to recreate the scenario observed in the bug where no resource as "/" added.

rombert: I am still interested in finding out how the unreadable root can take happen. Please describe the scenario in detail - either here, or in the Jira issue so that we can reproduce it in the Sling Starter. A content package that works with the Sling Starter would be ideal, but I can also work with "textual" descriptions.

If this becomes too complicated for this unit testing framework we can always add an integration test in https://github.com/apache/sling-org-apache-sling-launchpad-integration-tests .

But first, I want to understand the problem.

cziegeler:
https://issues.apache.org/jira/browse/SLING-10844 is already marked as resolved

PR #51, opened 2021-10-27
SLING-7975 support ordering of resources

kwin:
This depends upon https://github.com/apache/sling-org-apache-sling-api/pull/36


rombert (commented):
LGTM, but I'm not that knowledgeable in this area. It would be good IMO to add some tests for this functionality.

PR #52, opened 2021-11-04
SLING-10895 implement getPropertyMap()

joerghoh:


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -96,6 +98,8 @@
     private final ResourceResolverContext context;
 
     protected final Map<ResourceTypeInformation,Boolean> resourceTypeLookupCache = new ConcurrentHashMap<>();
+    
+    protected Map<String,Object> propertyMap;
```
rombert, line 102: Any particular reason to not make this private?

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -180,8 +184,25 @@ public void close() {
...
     }
+    
+    
+    protected void clearPropertyMap() {
+        if (propertyMap != null) {
+            for (Object value : propertyMap.values()) {
+                if (value instanceof Closeable) {
+                    try {
+                        ((Closeable) value).close();
+                    } catch (IOException e) {
```
rombert, line 198: I would catch `Exception` here. The reason is that if, for any reason, a consumer-supplied `close` method throws a `RuntimeException`, the cleaning is aborted and 

- unprocessed closeables will not be closed
- the resource resolver's close process is interrupted, leading to various leaks

kwin: I guess @rombert  meant not catch exceptions here and I agree. Let's not swallow those but propagate them through RR.close().

rombert: I actually meant to catch everything, as it can potentially do a lot of harm.

kwin: Then I disagree, just swallowing every exception without even logging calls for trouble. I would instead:
Catch the IOException, create a new exception object and put all those as suppressed exceptions. Then close the resource resolver and afterwards throw the newly created exception with all suppressed exceptions.

rombert: That sounds like a reasonable solution to me.

joerghoh: According to the API ResourceResolver.close does not throw an Exception.

https://sling.apache.org/apidocs/sling11/org/apache/sling/api/resource/ResourceResolver.html#close--

cziegeler: I think we either ignore the exceptions (but log them) or we remove the support for Closeable. I don't want that someone putting something random into that map blows up close()

joerghoh: I changed the code to log every exception on WARN, but to continue.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
joerghoh: fixed

joerghoh:
@cziegeler, @kwin : Thanks for your feedback and help!

PR #53, opened 2021-11-30
SLING-10945 add metrics

joerghoh:
Currently we are quite short in terms of metrics, which allow us to observe some basic numbers for the ResourceResolver. This PR adds 3 basic metrics:

* total number of aliases
* total number of vanity paths
* the number of unclosed resourceResolvers (since the last restart of the RR bundle)

The implementation is straight forward, there is a single ```ResourceResolverMetrics``` service, which either registers the Gauge services into OSGI or directly uses the Counter from Sling Commons Metrics. Basic test coverage for the ```ResourceResolverMetrics``` is provided, but I haven't augmented the existing test cases with validations, that their operations properly reflect in the metrics.

I tested this PR locally with my AEM instance, and all 3 metrics are reflected as JMX objects:
* org.apache.sling.resourceresolver:name=org.apache.sling.resourceresolve.unclosedResourceResolvers,type=Metrics
* org.apache.sling.resourceresolver:name=org.apache.sling.resourceresolve.numberOfVanityPaths,type=Metrics
* org.apache.sling.resourceresolver:name=org.apache.sling.resourceresolve.numberOfAliases,type=Metrics

as well as Metrics (flat in the sling namespace):
* sling:org.apache.sling.resourceresolver.numberOfAliases
* sling:org.apache.sling.resourceresolve.numberOfVanityPaths
* sling:org.apache.sling.resourceresolve.unclosedResourceResolvers

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -153,6 +153,7 @@
             <groupId>org.apache.sling</groupId>
             <artifactId>org.apache.sling.commons.metrics</artifactId>
             <version>1.2.8</version>
+            <scope>provided</scope>
         </dependency>
 
         <!-- For the Console Plugin of the ResourceResolverFactoryImpl -->
--- a/pom.xml
+++ b/pom.xml
@@ -149,6 +149,11 @@
             <artifactId>annotations</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.sling</groupId>
+            <artifactId>org.apache.sling.commons.metrics</artifactId>
+            <version>1.2.8</version>
+        </dependency>
 
         <!-- For the Console Plugin of the ResourceResolverFactoryImpl -->
         <dependency>
--- a/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
@@ -115,6 +115,7 @@ public class ResourceResolverFactoryActivator {
     @Reference
     ResourceAccessSecurityTracker resourceAccessSecurityTracker;
     
+    @SuppressWarnings("java:S3077")
     @Reference(cardinality = ReferenceCardinality.OPTIONAL, policyOption = ReferencePolicyOption.GREEDY, policy = ReferencePolicy.DYNAMIC)
     private volatile ResourceResolverMetrics metrics;
 
--- a/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
@@ -51,6 +51,7 @@ import org.osgi.service.component.annotations.Modified;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.osgi.service.component.annotations.ReferencePolicy;
+import org.osgi.service.component.annotations.ReferencePolicyOption;
 import org.osgi.service.event.EventAdmin;
 import org.osgi.service.metatype.annotations.Designate;
 import org.slf4j.Logger;
@@ -114,7 +115,7 @@ public class ResourceResolverFactoryActivator {
     @Reference
     ResourceAccessSecurityTracker resourceAccessSecurityTracker;
     
-    @Reference
+    @Reference(cardinality = ReferenceCardinality.OPTIONAL, policyOption = ReferencePolicyOption.GREEDY)
     private ResourceResolverMetrics metrics;
 
     volatile ResourceProviderTracker resourceProviderTracker;
@@ -162,8 +163,8 @@ public class ResourceResolverFactoryActivator {
         return stringInterpolationProvider;
     }
     
-    public ResourceResolverMetrics getResourceResolverMetrics() {
-        return this.metrics;
+    public Optional<ResourceResolverMetrics> getResourceResolverMetrics() {
+        return Optional.ofNullable(this.metrics);
     }
 
     /**
--- a/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
@@ -115,8 +115,8 @@ public class ResourceResolverFactoryActivator {
     @Reference
     ResourceAccessSecurityTracker resourceAccessSecurityTracker;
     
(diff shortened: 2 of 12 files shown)


joerghoh:
(prefixed the metric names with 'org.apache.sling.resourceresolver' with a new PR, description has been adjusted)

src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
```diff
@@ -94,6 +95,7 @@ public ResourceMapperImplTest(boolean optimiseAliasResolution) {
     @Before
     public void prepare() throws LoginException {
 
+        ctx.registerService(ResourceResolverMetrics.class, Mockito.mock(ResourceResolverMetrics.class));
```
rombert, line 98: I think we shouldn't mix Sling mocks and mockito here. Can you just pass the plain object?

pom.xml
```diff
@@ -149,6 +149,11 @@
             <artifactId>annotations</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
rombert, line 152: Can we make this dependency optional? I am worried that this will lead in a wave of dependency breakages in various consumer projects.

pom.xml
```diff
@@ -149,6 +149,11 @@
...
+            <groupId>org.apache.sling</groupId>
+            <artifactId>org.apache.sling.commons.metrics</artifactId>
+            <version>1.2.8</version>
```
rombert, line 155: The scope should be provided.

rombert (commented):
LGTM overall, a couple of review notes to clarify.

pom.xml
joerghoh: Instead of passing around a nullable value, I used ```java.util.Optional```.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.sling.commons.metrics.MetricsService;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceRegistration;
+import org.osgi.service.component.ComponentContext;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
+import org.osgi.service.component.annotations.Reference;
+
+@Component(service=ResourceResolverMetrics.class, immediate=true)
```
cziegeler, line 36: I don't see a need for "immedaite"

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
```diff
@@ -0,0 +1,122 @@
...
+    
+    // number of aliases
+    private ServiceRegistration<Gauge<Long>> numberOfAliasesGauge;
+    private Supplier<Long> numberOfAliasesSupplier = ZERO_SUPPLIER;
+    
+    private Counter unclosedResourceResolvers;
+    
+    
+    @Activate
+    protected void activate(ComponentContext context) {
```
cziegeler, line 58: Why ComponentContext, isn't BundleContext enough?

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
```diff
@@ -0,0 +1,122 @@
...
+        BundleContext bundleContext = context.getBundleContext();
+        numberOfVanityPathsGauge = registerGauge(bundleContext, METRICS_PREFIX + ".numberOfVanityPaths", () -> numberOfVanityPathsSupplier );
```
cziegeler, line 60: Can we avoid dealing with framework api?

joerghoh: There is no equivalent of ```metricsService.counter()``` for Gauge :-/ , but only https://issues.apache.org/jira/browse/SLING-5966

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
joerghoh: removed immediate

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
joerghoh: fixed

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -113,6 +114,9 @@
 
     @Reference
     ResourceAccessSecurityTracker resourceAccessSecurityTracker;
+    
+    @Reference(cardinality = ReferenceCardinality.OPTIONAL, policyOption = ReferencePolicyOption.GREEDY)
+    private ResourceResolverMetrics metrics;
```
cziegeler, line 119: This reference should also be dynamic, this allows that the field is updated when the service appears/disappears instead of restarting the component. The field should be volatile

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
cziegeler: Right, but the two gauge services could be DS components with a reference to the ResourceResolverMetrics

joerghoh: From my point of view it would get more complex and require more code, and I don't see any benefit doing so.

cziegeler: Mixin DS and framework api is an anti pattern; while I don't agree that using DS is substantially more code or more complex, I guess we can leave it like this for now

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
joerghoh: done

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
joerghoh: I raised https://issues.apache.org/jira/browse/SLING-10967 to move this logic into Sling Commons Metrics. I will keep this logic here until the change in the Sling Commons Metric is implemented.

joerghoh:
Thank you for the review!

PR #54, opened 2021-12-09
SLING-10958 send only 1 event even if multiple vanity path updates happen

joerghoh:


cziegeler (approved):
lgtm, thanks

joerghoh:
thanks @cziegeler

joerghoh:
For the record: The SLING ID in this PR is incorrect :-( It should be SLING-10968

PR #57, opened 2022-01-26
SLING-11092: ResourceResolver - update org.apache.sling.testing.osgi-mock test dependency

reschke:


rombert:
Thanks for the contribution!

PR #60, opened 2022-02-03
SLING-11113: remove bloom filter persistence

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -18,34 +18,6 @@
  */
 package org.apache.sling.resourceresolver.impl.mapping;
 
-import org.apache.commons.lang3.StringUtils;
```
cziegeler, line 21: The PR is reordering imports - it would be good to avoid this to see the real changes

reschke: will do

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -146,14 +138,10 @@
 
     private final AtomicLong vanityCounter;
 
-    private final File vanityBloomFilterFile;
-
     private byte[] vanityBloomFilter;
 
     private Timer timer;
```
cziegeler, line 143: Is the timer still needed?

reschke: good q

cziegeler (approved):
Lgtm, thanks

reschke:
Thanks!

PR #63, opened 2022-05-05
SLING-11296: streamline query code and add timing to logging

reschke:


joerghoh:
When you capture the timing information, can you expose this information as a metric as well?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -975,52 +976,52 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
     private Map<String, Map<String, String>> loadAliases(final ResourceResolver resolver) {
         final Map<String, Map<String, String>> map = new ConcurrentHashMap<>();
-        final String queryString = updateAliasQuery();
+        final String queryString = generateAliasQuery();
+
         log.debug("start alias query: {}", queryString);
+        long queryStart = System.nanoTime();
         final Iterator<Resource> i = resolver.findResources(queryString, "sql");
-        log.debug("end alias query");
+        long queryElapsed = System.nanoTime() - queryStart;
```
joerghoh, line 989: I am not sure how relevant this information is, because the query itself is executed lazily (IIRC it just loads 20 hits on execution), and just fetching the additional result rows gets costly.

I would drop this information and just collect the "processElapsed" value you capture below.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -975,52 +976,52 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
         baseQuery.append(" ").append("WHERE");
 
-        if(allowedLocations.isEmpty()){
+        if (allowedLocations.isEmpty()) {
             String jcrSystemPath = StringUtils.removeEnd(JCR_SYSTEM_PREFIX, "/");
-            baseQuery.append(" ").append("(").append("NOT ISDESCENDANTNODE(page,")
-                    .append("\"").append(jcrSystemPath).append("\"").append("))");
-
-        }else{
+            baseQuery.append(" (").append("NOT ISDESCENDANTNODE(page,'").append(queryLiteral(jcrSystemPath)).append("'))");
```
joerghoh, line 1015: would it make sense to use String.format or something else? I think it's not really readable what's happening here.

reschke:
Good question; in general, these queries run only once on startup - does this qualify for metrics capturing?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: Well, it's more readable than before; I didn't want to change too much at once.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: Does it hurt? Reminder: this is for people looking at DEBUG logs.

joerghoh:
You never have the loglevel on DEBUG when you need it :-) Assuming that this is just needed during development and validation, we don't need the exact timing, especially as it correlates somehow linearly to the total number of aliases which we already capture as a metric.

reschke:
Well. We *assume* that it collerates linearly. We don't know for sure. We also don't have the absolute numbers (for timing), as this also depends on persistence type, network connectivity, database load etc. When a customer says "it takes to long to startup", I want to be able to see what exactly took how long.

Yes, that's a different use case than monitoring running systems, where metrics make a lot of sense.

cziegeler (approved):
Lgtm, thanks

reschke:
thx

PR #66, opened 2022-05-23
Sling 11331: resource resolver: events need to be queued until vanity path loading is complete

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -139,6 +143,8 @@ public class MapEntries implements
 
     private byte[] vanityBloomFilter;
 
+    private boolean vanityPathsProcessed;
```
cziegeler, line 146: I think this should be volatile, or maybe use AtomicBoolean

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -714,6 +783,17 @@ public void onChange(final List<ResourceChange> changes) {
                 continue;
             }
 
+            // during startup: just enqueue the events
```
cziegeler, line 786: It's maybe better to do the check at method start and then handle all changes in the same way. If the check is done here, the first changes might go into the queue, while the remaining ones are not.

reschke: that could only happen when "vanityPathsProcessed" ever changes from true to false, right?

cziegeler: right

cziegeler (approved):
Lgtm., thanks

PR #69, opened 2022-05-25
SLING-11328: add diagnostics when paths are duplicates

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1077,25 +1077,28 @@ private void gather(final ResourceResolver resolver, final List<MapEntry> entrie
...
-            List<MapEntry> entriesCopy =new ArrayList<>(entries);
-            entriesCopy.add(entry);
-            // and finally sort list
-            Collections.sort( entriesCopy);
-            entryMap.put(key, entriesCopy);
+            List<MapEntry> entries = entryMap.get(key);
+            if (entries == null) {
+                entries = new ArrayList<>();
+                entries.add(entry);
+                entryMap.put(key, Collections.synchronizedList(entries));
```
cziegeler, line 1088: With a synchronized list you punish all read calls which outweight the write calls. Thats why the code is creating a new list with every modification. I guess these days a copy on write list could be used instead. And instead of sorting, the insert could search for the right index to insert into (traversing the list once)

reschke: Good point. Will paddle back and just leave in the log messages.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1077,25 +1077,30 @@ private void gather(final ResourceResolver resolver, final List<MapEntry> entrie
...
-        }
-
-        List<MapEntry> entries = entryMap.get(key);
-        if (entries == null) {
-            entries = new ArrayList<>();
-            entries.add(entry);
-            // and finally sort list
-            Collections.sort(entries);
-            entryMap.put(key, entries);
         } else {
```
cziegeler, line 1083: When entry is null, the return statement will be executed. So no need to do a "else" here

reschke: The intent was to have both "returns" at the same level. Gimme a sec.

cziegeler (approved):

PR #70, opened 2022-05-30
SLING-11343: add metrics related to vanity path lookup and bloom filter behavior

reschke:


reschke:
This PS might be introducing a bug when the cache size is unlimited, and BG init is enabled. Working on it...

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -709,10 +718,38 @@ public Map<String, List<String>> getVanityPathMappings() {
...
-        if (!vanityPathsProcessed.get() || BloomFilterUtils.probablyContains(vanityBloomFilter, vanityPath)) {
+        boolean initFinished = vanityPathsProcessed.get();
+        boolean probablyPresent = false;
+
+        if (initFinished) {
+            // total number of lookups after init (and when cache not complete)
+            long current = this.vanityPathLookups.incrementAndGet();
+            if (current >= Long.MAX_VALUE - 100000) {
+                // reset counters when we get close the limit
+                synchronized (this) {
```
cziegeler, line 729: I don't think we need synchronized here; its not protecting the access anyway

cziegeler (approved):

reschke:
thx.

PR #73, opened 2022-06-13
[POC] Run the Sling Starter ITs with the ResourceResolver bundle

rombert:
Will fail, needs SNAPSHOTs with the changes from https://github.com/apache/sling-parent/pull/27 and https://github.com/apache/sling-feature-launcher-maven-plugin/pull/8


rombert:
The PR fails because the SNAPSHOT parent can't be found ( chicken and egg problem ).

PR #74, opened 2022-06-24
SLING-11394: improve diagnostics when # of vanity paths is close or greater than configured cache size

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1251,20 +1251,29 @@ private Map<String, List<String>> loadVanityPaths(ResourceResolver resolver) {
...
                         || vanityCounter.longValue() < this.factory.getMaxCachedVanityPathEntries();
                 loadVanityPath(resource, resolveMapsMap, targetPaths, addToCache);
             }
         }
         long processElapsed = System.nanoTime() - processStart;
-        log.debug("processed {} vanityPaths in {}ms", count, TimeUnit.NANOSECONDS.toMillis(processElapsed));
+        log.debug("processed {} vanityPaths (of which {} in scope) in {}ms", count, countInScope, TimeUnit.NANOSECONDS.toMillis(processElapsed));
+        if (!isAllVanityPathEntriesCached()) {
+            if (countInScope > this.factory.getMaxCachedVanityPathEntries()) {
+                log.warn("Number of vanity paths in scope ({}) exceeds configured cache size ({})", countInScope, this.factory.getMaxCachedVanityPathEntries());
```
joerghoh, line 1272: What is the consequence of this? 
Also can we add a call-to-action to this? Would "Increase the cache size" be the right recommendation?

joerghoh (approved):

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: The consequence is that request for certainl paths will be slow because they'll hit the storage (by query). The recommendation would be to either reduce the number of vanity paths, or increase the cache size.

Do you want to make a proposal for the concrete text?

joerghoh:
```suggestion
                log.warn("Number of vanity paths in scope ({}) exceeds configured cache size ({}); this and any following vanityPath will be much slower. Consider increasing the cache size or decrease the number of vanity paths", countInScope, this.factory.getMaxCachedVanityPathEntries());
```

cziegeler (approved):

PR #78, opened 2022-08-05
Various improvements for the webconsole plugin

rombert:


cziegeler (approved):
lgtm, thanks

src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java
```diff
@@ -160,14 +169,22 @@ protected void doGet(final HttpServletRequest request,
...
+        pw.println("<form method='post'>");
+        pw.print("Test ");
         pw.print("<input type='text' name='" + ATTR_TEST + "' value='");
         if (test != null) {
             pw.print(ResponseUtil.escapeXml(test));
         }
-        pw.println("' class='input' size='50'>");
+        pw.println("' class='input' size='20'>");
+        pw.print("User (optional)");
+        pw.print("<input type='text' name='" + ATTR_USER + "' value='");
```
kwin, line 181: Maybe add a hint that this does only work/has an effect on resources resolved by the JCR resource provider.

src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java
```diff
@@ -255,19 +278,42 @@ protected void doPost(HttpServletRequest request,
...
         }
         response.sendRedirect(redirectTo);
     }
 
+    private ResourceResolver getImpersonatedResourceResolver(HttpServletRequest request, final String user)
+            throws LoginException {
+
+        // resolver is set by the auth.core bundle in case of successful authentication, so it should
+        // always be there
+        Object resolverAttribute = request.getAttribute(AuthenticationSupport.REQUEST_ATTRIBUTE_RESOLVER);
```
kwin, line 299: Shouldn't this rather be based on service resource resolver as well to make it work with all web console security providers?

src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java
```diff
@@ -255,19 +278,42 @@ protected void doPost(HttpServletRequest request,
...
+        if ( !(resolverAttribute instanceof ResourceResolver) ) {
+            throw new IllegalArgumentException("No " + ResourceResolver.class.getSimpleName() + " found in request, unable to proceed with impersonation");
+        }
+
+        @SuppressWarnings("resource") // not a leak, we don't own this resolver
+        ResourceResolver currentResolver = (ResourceResolver) resolverAttribute;
+
+        Map<String, Object> authenticationInfo = new HashMap<>();
+        authenticationInfo.put(ResourceResolverFactory.USER_IMPERSONATION, user);
+        authenticationInfo.put(JcrResourceConstants.AUTHENTICATION_INFO_SESSION, currentResolver.adaptTo(Session.class));
```
kwin, line 309: This adaptTo may return null in case JCR resource provider is not used, a null check is necessary here.

kwin (changes requested):
LGTM overall, just some minor remarks to better handle some edge cases.

src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java
cziegeler: The import of JCR packages is optional, I assume the above code does not work if the JCR bundle is not installed?
We should keep that dependency as optional and not require JCR api to be installed

rombert: There are two issues to discuss here

1. The javax.jcr import is already optional - https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/78/files#diff-6d1033f3cc4642b36cd395cf203fdc2d963f51e46cb1949508765a9756743c45R2 . I _think_ that a `NoClassDefFoundError` will be thrown from the method, then caught at https://github.com/apache/sling-org-apache-sling-resourceresolver/blob/11f26ad706a350269e27ca42a2cbcf22a4724ce1/src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java#L265 . This should not affect the path where impersonation is not used. Do we need to handle this in some other way, e.g. using the bundle class loader to dynamically load the class definition?
2. The `adaptTo` call can return null nonetheless and I will take it into acount

src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java
rombert: Will do.

src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java
rombert: Do you mean that we should try to create an additional mapping for the console using a service user that has impersonation rights?

kwin: Just try to get a resourceresolver based on the service one. The one returned as request attribute might not have the according rights

src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java
cziegeler: My bad, I didn't read the full code. You are right, that this will work. And just to be sure I gave it a test run in a sling application that does not have the jcr api and except for impersonation, everything works

src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java
```diff
@@ -255,19 +278,42 @@ protected void doPost(HttpServletRequest request,
...
     }
 
+    private ResourceResolver getImpersonatedResourceResolver(HttpServletRequest request, final String user)
+            throws LoginException {
+
+        // resolver is set by the auth.core bundle in case of successful authentication, so it should
+        // always be there
+        Object resolverAttribute = request.getAttribute(AuthenticationSupport.REQUEST_ATTRIBUTE_RESOLVER);
+        if ( !(resolverAttribute instanceof ResourceResolver) ) {
+            throw new IllegalArgumentException("No " + ResourceResolver.class.getSimpleName() + " found in request, unable to proceed with impersonation");
```
cziegeler, line 301: It is a common case if the web console is not using the Sling authentication provider that the request attribute is not set. In that case, it currently displays:
Test Failure: java.lang.IllegalArgumentException: No ResourceResolver found in request, unable to proceed with impersonation
I think it would be could if just the message without the exception type would be displayed

src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java
rombert: If I try to set up impersonation based on the existing resolver ( https://github.com/apache/sling-org-apache-sling-resourceresolver/blob/11f26ad706a350269e27ca42a2cbcf22a4724ce1/src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java#L246 ) I get back 

> Test Failure: org.apache.sling.api.resource.LoginException: Impersonation not allowed.

kwin: You need to adjust privileges of the underlying technical user.

rombert: Looking at the Oak implementation, I see that impersonation works if either:
- the impersonator is an admin
- the impersonator is included in the `rep:impersonators` property of the impersonated user

https://github.com/apache/jackrabbit-oak/blob/a90566744551246535f65c2aefc5a44fd5275490/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/ImpersonationImpl.java#L125-L146

I am not sure if either of these is possible or desireable for a service user. Do you see another way?

kwin: Actually the user needs to be "admin", just being member of the administrators group is IMHO not enough. I don't think that there is an option yet for a user to enable him to impersonate as anyone else. Might be a good extension though. 
The same limitation applies to the user doing the webconsole request (in case the Sling Webconsole Security provider is used), so in fact this option does only work for admin with all other users.
Therefore I would suggest to use a new administrative resource resolver with impersonation and whitelist the usage accordingly.

src/main/java/org/apache/sling/resourceresolver/impl/console/ResourceResolverWebConsolePlugin.java
rombert: @kwin suggested that we use an admin resolver instead (and include the bundle in the allow list ). If we would stop looking up the ResourceResolver in the request attribute, would it solve this issue?

https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/78#discussion_r941353233

cziegeler: I'm not really a fan of adding another case for using an admin resource resolver, especially not in the web console

PR #81, opened 2022-08-15
SLING-11541: vanity path query: attempt to query sorted by first vanity path, check results

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1275,25 +1276,40 @@ private static boolean isAliasValid(String alias) {
         return invalid;
     }
 
+    private Iterator<Resource> queryAllVanityPaths(String query) {
+        log.debug("start vanityPath query: {}", query);
+        long queryStart = System.nanoTime();
+        final Iterator<Resource> i = resolver.findResources(query, "JCR-SQL2");
+        long queryElapsed = System.nanoTime() - queryStart;
```
thomasmueller, line 1283: This will probably not do what you want, because JCR queries are processed lazily (while iterating, not when executed). So, I would not log at this stage. Instead, I would measure the time on a higher level, while iterating over the result.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1275,25 +1276,40 @@ private static boolean isAliasValid(String alias) {
...
                 + " AND [sling:vanityPath] IS NOT NULL";
+        final String queryStringWithSort = baseQueryString + " ORDER BY FIRST([sling:vanityPath]), [jcr:path]";
 
-        log.debug("start vanityPath query: {}", queryString);
-        long queryStart = System.nanoTime();
-        final Iterator<Resource> i = resolver.findResources(queryString, "JCR-SQL2");
-        long queryElapsed = System.nanoTime() - queryStart;
-        log.debug("end vanityPath query; elapsed {}ms", TimeUnit.NANOSECONDS.toMillis(queryElapsed));
+        boolean supportsSort = true;
+        Iterator<Resource> i;
```
thomasmueller, line 1300: Nitpicking: "i" is typically used for "int i"; I would use "it" for iterators.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1303,7 +1319,13 @@ private Map<String, List<String>> loadVanityPaths(ResourceResolver resolver) {
                 countInScope += 1;
                 final boolean addToCache = isAllVanityPathEntriesCached()
                         || vanityCounter.longValue() < this.factory.getMaxCachedVanityPathEntries();
-                loadVanityPath(resource, resolveMapsMap, targetPaths, addToCache);
+                String firstVanityPath = loadVanityPath(resource, resolveMapsMap, targetPaths, addToCache);
+                if (supportsSort && firstVanityPath != null) {
+                    if (previousVanityPath != null && firstVanityPath.compareTo(previousVanityPath) < 0) {
+                        log.error("Sorting by first(vanityPath) does not appear to work; got " + firstVanityPath + " after " + previousVanityPath);
```
thomasmueller, line 1325: It might be that sorting on the server and sorting on the client doesn't match for some edge cases (like, client locale doesn't match server locale).

For this reason, I would recommend to log a warning, not an error.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/InMemoryResourceProvider.java
```diff
@@ -103,7 +103,7 @@ public Iterator<Resource> findResources(@NotNull ResolveContext<Void> ctx, Strin
                         .iterator();
                 }
 
-                if ( "JCR-SQL2".equals(language) && "SELECT [sling:vanityPath], [sling:redirect], [sling:redirectStatus] FROM [nt:base] WHERE NOT isdescendantnode('/jcr:system') AND [sling:vanityPath] IS NOT NULL".equals(query) ) {
+                if ( "JCR-SQL2".equals(language) && "SELECT [sling:vanityPath], [sling:redirect], [sling:redirectStatus] FROM [nt:base] WHERE NOT isdescendantnode('/jcr:system') AND [sling:vanityPath] IS NOT NULL ORDER BY FIRST([sling:vanityPath]), [jcr:path]".equals(query) ) {
```
thomasmueller, line 106: This query doesn't use keyset pagination currently, so it can still fail when there are many entries.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: ack

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: We do that as well (and have been doing so...)

src/test/java/org/apache/sling/resourceresolver/impl/mapping/InMemoryResourceProvider.java
reschke: Yes, we'll get to that next.

reschke:
@cziegeler - I think this is ready to go.

PR #82, opened 2022-09-15
SLING-11581: use keyset pagination for vanity path query

reschke:


src/test/java/org/apache/sling/resourceresolver/impl/mapping/InMemoryResourceProvider.java
```diff
@@ -103,7 +103,7 @@ public Iterator<Resource> findResources(@NotNull ResolveContext<Void> ctx, Strin
                         .iterator();
                 }
 
-                if ( "JCR-SQL2".equals(language) && "SELECT [sling:vanityPath], [sling:redirect], [sling:redirectStatus] FROM [nt:base] WHERE NOT isdescendantnode('/jcr:system') AND [sling:vanityPath] IS NOT NULL ORDER BY FIRST([sling:vanityPath]), [jcr:path]".equals(query) ) {
+                if ( "JCR-SQL2".equals(language) && "SELECT [sling:vanityPath], [sling:redirect], [sling:redirectStatus] FROM [nt:base] WHERE NOT isdescendantnode('/jcr:system') AND [sling:vanityPath] IS NOT NULL AND FIRST([sling:vanityPath]) > '' ORDER BY FIRST([sling:vanityPath])".equals(query) ) {
```
joerghoh, line 106: Would it make sense to move this query into a constant (or build a function returning it if it needs parameters), which is used in both production code and the tests?

reschke: Likely, as the unit tests really depend on that implementation detail.

I would propose to do this in a separate PR though, as it affects more places.

joerghoh (approved):

reschke:
will look at improving test coverage...

joerghoh (approved):

PR #85, opened 2022-09-30
SLING-11604: Async VanityPathInitializer should log when completed

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -329,7 +329,8 @@ private void execute() {
             try (ResourceResolver resolver = factory
                     .getServiceResourceResolver(factory.getServiceUserAuthenticationInfo("mapping"))) {
 
-                log.debug("vanity path init - start");
+                long initStart = System.nanoTime();
+                log.info("vanity path initialization - start");
```
joerghoh, line 333: I don't think that this message is really helpful on INFO level, because it will be always started along the SlingRepository.
Leaving on DEBUG would be fine for me.

joerghoh (approved):

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: Ack.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -341,7 +342,8 @@ private void execute() {
                 // drain once more in case more events have arrived
                 drainQueue(resourceChangeQueue);
 
-                log.debug("vanity path init - end");
+                long initElapsed = System.nanoTime() - initStart;
+                log.info("vanity path initialization - end, elapsed {}ms", TimeUnit.NANOSECONDS.toMillis(initElapsed));
```
cziegeler, line 346: I think we should have some consistency in the log message format. Below when the exception is logged it is "VanityPath", here (and above) it is "vanity path"; I haven't checked the whole code - but at least here using "VanityPath" seems to be more consistent

PR #89, opened 2023-01-09
[SLING-11742] Provide alternative equitable terminology for properties

cristic83:
This PR introduces alternative terminology for terms that are considered un-appropriate for vanity paths, as described in [1]. The approach taken was to provide additional properties to the existing properties so that backwards compatibility is kept. For the sake of backwards compatibility, the existing vanity paths properties are checked first and if these are not defined, the newly introduced properties are checked. A log message is produced in this case.

[1] https://issues.apache.org/jira/browse/SLING-11742

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
@@ -81,46 +81,6 @@ public class ResourceResolverFactoryActivator {
         public volatile CommonResourceResolverFactoryImpl commonFactory;
     }
 
-    private static final class VanityPathConfigurer {
-        private final Logger logger = LoggerFactory.getLogger(this.getClass());
-
-        void configureVanityPathPrefixes(String[] pathPrefixes, String[] pathPrefixesFallback,
-                                         String pathPrefixesPropertyName, String pathPrefixesFallbackPropertyName,
-                                         Consumer<String[]> filteredPathPrefixesConsumer) {
-            if (pathPrefixes != null) {
-                configureVanityPathPrefixes(pathPrefixes, filteredPathPrefixesConsumer);
-            } else {
-                logger.debug("The " + pathPrefixesPropertyName + " was null. Using the " +
-                    pathPrefixesFallbackPropertyName + " instead if defined.");
-                if (pathPrefixesFallback != null) {
-                    configureVanityPathPrefixes(pathPrefixesFallback, filteredPathPrefixesConsumer);
-                }
-            }
-        }
-
-        private static void configureVanityPathPrefixes(String[] pathPrefixes, Consumer<String[]> pathPrefixesConsumer) {
-            final List<String> filterVanityPaths = filterVanityPathPrefixes(pathPrefixes);
-            if (filterVanityPaths.size() > 0) {
-                pathPrefixesConsumer.accept(filterVanityPaths.toArray(new String[filterVanityPaths.size()]));
-            }
-        }
-
-        @NotNull
-        private static List<String> filterVanityPathPrefixes(String[] vanityPathPrefixes) {
-            final List<String> prefixList = new ArrayList<>();
-            for (final String value : vanityPathPrefixes) {
-                if (value.trim().length() > 0) {
-                    if (value.trim().endsWith("/")) {
-                        prefixList.add(value.trim());
-                    } else {
-                        prefixList.add(value.trim() + "/");
-                    }
-                }
-            }
-            return prefixList;
-        }
-    }
-
     /** Logger. */
     private final Logger logger = LoggerFactory.getLogger(this.getClass());
 
@@ -385,16 +345,16 @@ public class ResourceResolverFactoryActivator {
         // vanity path white list
         this.vanityPathWhiteList = null;
         vanityPathConfigurer.configureVanityPathPrefixes(config.resource_resolver_vanitypath_whitelist(),
-            config.resource_resolver_vanitypath_allowedlist(),
+            config.resource_resolver_vanitypath_allowlist(),
             "resource_resolver_vanitypath_whitelist",
-            "resource_resolver_vanitypath_allowedlist",
+            "resource_resolver_vanitypath_allowlist",
             filteredPrefixes -> this.vanityPathWhiteList = filteredPrefixes);
         // vanity path black list
         this.vanityPathBlackList = null;
         vanityPathConfigurer.configureVanityPathPrefixes(config.resource_resolver_vanitypath_blacklist(),
-            config.resource_resolver_vanitypath_deniedlist(),
+            config.resource_resolver_vanitypath_denylist(),
             "resource_resolver_vanitypath_blacklist",
-            "resource_resolver_vanitypath_deniedlist",
+            "resource_resolver_vanitypath_denylist",
             filteredPrefixes -> this.vanityPathBlackList = filteredPrefixes);
 
         // check for required property
--- a/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
@@ -24,6 +24,7 @@ import java.lang.reflect.Proxy;
 import java.util.*;
 
 
+import java.util.function.Consumer;
 import org.apache.commons.collections4.BidiMap;
 import org.apache.commons.collections4.bidimap.TreeBidiMap;
 import org.apache.commons.lang3.StringUtils;
@@ -39,6 +40,7 @@ import org.apache.sling.resourceresolver.impl.providers.ResourceProviderTracker;
 import org.apache.sling.resourceresolver.impl.providers.ResourceProviderTracker.ChangeListener;
 import org.apache.sling.resourceresolver.impl.providers.RuntimeServiceImpl;
 import org.apache.sling.serviceusermapping.ServiceUserMapper;
+import org.jetbrains.annotations.NotNull;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.Constants;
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
```diff
@@ -163,15 +163,27 @@
...
                     "such a list is configured, only vanity paths from resources starting with this prefix " +
-                    " are considered. If the list is empty, all vanity paths are used.")
+                    " are considered. If the list is empty, we fallback to resource_resolver_vanitypath_allowedlist.")
     String[] resource_resolver_vanitypath_whitelist();
 
+    @AttributeDefinition(name = "Allowed Vanity Path Location",
+        description ="This setting can contain a list of path prefixes, e.g. /libs/, /content/. If " +
+            "such a list is configured, only vanity paths from resources starting with this prefix " +
+            " are considered. If the list is empty, all vanity paths are used.")
+    String[] resource_resolver_vanitypath_allowedlist();
```
rombert, line 173: Please use 'allowlist' instead of 'allowedlist' . See also https://cwiki.apache.org/confluence/display/SLING/Removal+of+problematic+language

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
```diff
@@ -163,15 +163,27 @@
...
                     "such a list is configured,vanity paths from resources starting with this prefix " +
-                    " are not considered. If the list is empty, all vanity paths are used.")
+                    " are not considered. If the list is empty, we fallback to resource_resolver_vanitypath_deniedlist.")
     String[] resource_resolver_vanitypath_blacklist();
 
+    @AttributeDefinition(name = "Denied Vanity Path Location",
+        description ="This setting can contain a list of path prefixes, e.g. /misc/. If " +
+            "such a list is configured,vanity paths from resources starting with this prefix " +
+            " are not considered. If the list is empty, all vanity paths are used.")
+    String[] resource_resolver_vanitypath_deniedlist();
```
rombert, line 185: Please use 'denylist' instead of 'deniedlist'

src/test/java/org/apache/sling/resourceresolver/impl/MockedResourceResolverImplTest.java
```diff
@@ -195,6 +195,11 @@ public String[] resource_resolver_virtual() {
 
             @Override
             public String[] resource_resolver_vanitypath_whitelist() {
+                return resource_resolver_vanitypath_allowedlist();
```
rombert, line 198: This can get confusing, please return just `null`.

src/test/java/org/apache/sling/resourceresolver/impl/MockedResourceResolverImplTest.java
```diff
@@ -215,6 +220,11 @@ public int resource_resolver_vanitypath_bloomfilter_maxBytes() {
 
             @Override
             public String[] resource_resolver_vanitypath_blacklist() {
+                return resource_resolver_vanitypath_deniedlist();
```
rombert, line 223: This can get confusing, please return just `null`.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -79,6 +81,45 @@ private static final class FactoryRegistration {
         public volatile CommonResourceResolverFactoryImpl commonFactory;
     }
 
+    private static final class VanityPathConfigurer {
```
rombert, line 84: Please extract a test for this class, having the logic separated makes it easy.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -79,6 +81,45 @@ private static final class FactoryRegistration {
...
+        private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+        void configureVanityPathPrefixes(String[] pathPrefixes, String[] pathPrefixesFallback,
+                                         String pathPrefixesPropertyName, String pathPrefixesFallbackPropertyName,
+                                         Consumer<String[]> filteredPathPrefixesConsumer) {
+            if (pathPrefixes != null) {
+                configureVanityPathPrefixes(pathPrefixes, filteredPathPrefixesConsumer);
```
rombert, line 91: Please log a WARN in the case where both the regular prefixes and the fallback ones are defined; this is probably a customer error. We should rely on the old behaviour in this case  and only take information from `config.resource_resolver_vanitypath_whitelist()`.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -79,6 +81,45 @@ private static final class FactoryRegistration {
         public volatile CommonResourceResolverFactoryImpl commonFactory;
     }
 
+    private static final class VanityPathConfigurer {
+        private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+        void configureVanityPathPrefixes(String[] pathPrefixes, String[] pathPrefixesFallback,
```
rombert, line 87: I wonder if it makes sense to pass in the whole config object, so simplify the signature ; now we pass to fields and their names, passing in the object might make things nicer.

rombert (changes requested):
Thanks for the PR @cristic83 ! I've addede a few comments.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
cristic83: I attempted to implement this suggestion and the result looks like below. I don't think it's all that nicer,what do you think?
`void configureVanityPAthPrefix(ResourceResolverFactoryConfig config, Consumer<String[]> allowVanityPathConsumer,  
                                   Consumer<String[]> denyVanityPathConsumer) {
        configureVanityPathPrefixes(config.resource_resolver_vanitypath_whitelist(),
            config.resource_resolver_vanitypath_allowlist(),
            "resource_resolver_vanitypath_whitelist",
            "resource_resolver_vanitypath_allowlist",
            allowVanityPathConsumer);
        configureVanityPathPrefixes(config.resource_resolver_vanitypath_blacklist(),
            config.resource_resolver_vanitypath_denylist(),
            "resource_resolver_vanitypath_blacklist",
            "resource_resolver_vanitypath_denylist",
            denyVanityPathConsumer);
    }`

rombert:
I've assigned myself as a reviewer to make sure it's on my TODO list. Not sure when I will get around to it, if anyone beats me to it I won't mind :-)

cziegeler:
@cristic83 I suggest we take a similar approach to https://github.com/apache/sling-org-apache-sling-jcr-base/pull/9 where we have let the code just use the new properties, but have a conversion from old to new

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
```diff
@@ -163,15 +163,27 @@
     @AttributeDefinition(name = "Allowed Vanity Path Location",
         description ="This setting can contain a list of path prefixes, e.g. /libs/, /content/. If " +
                     "such a list is configured, only vanity paths from resources starting with this prefix " +
-                    " are considered. If the list is empty, all vanity paths are used.")
+                    " are considered. If the list is empty, we fallback to resource.resolver.vanitypath.allowlist.")
```
cziegeler, line 166: I think the fallback should be in the opposite direction: if the new property is not used, the old one will be used

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
```diff
@@ -163,15 +163,27 @@
...
+        description ="This setting can contain a list of path prefixes, e.g. /libs/, /content/. If " +
+            "such a list is configured, only vanity paths from resources starting with this prefix " +
+            " are considered. If the list is empty, all vanity paths are used.")
+    String[] resource_resolver_vanitypath_allowlist();
+
     @AttributeDefinition(name = "Denied Vanity Path Location",
         description ="This setting can contain a list of path prefixes, e.g. /misc/. If " +
                     "such a list is configured,vanity paths from resources starting with this prefix " +
-                    " are not considered. If the list is empty, all vanity paths are used.")
+                    " are not considered. If the list is empty, we fallback to resource.resolver.vanitypath.denylist.")
```
cziegeler, line 178: same as above

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
```diff
@@ -163,15 +163,27 @@
...
+                    " are considered. If the list is empty, we fallback to resource.resolver.vanitypath.allowlist.")
     String[] resource_resolver_vanitypath_whitelist();
 
+    @AttributeDefinition(name = "Allowed Vanity Path Location",
+        description ="This setting can contain a list of path prefixes, e.g. /libs/, /content/. If " +
+            "such a list is configured, only vanity paths from resources starting with this prefix " +
+            " are considered. If the list is empty, all vanity paths are used.")
+    String[] resource_resolver_vanitypath_allowlist();
+
     @AttributeDefinition(name = "Denied Vanity Path Location",
```
cziegeler, line 175: We should remove the metatype information for this property

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
```diff
@@ -163,15 +163,27 @@
     @AttributeDefinition(name = "Allowed Vanity Path Location",
```
cziegeler, line 163: we should remove the metatype info for this property

src/main/java/org/apache/sling/resourceresolver/impl/VanityPathConfigurer.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class VanityPathConfigurer {
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    void configureVanityPathPrefixes(String[] pathPrefixes, String[] pathPrefixesFallback,
```
cziegeler, line 31: This could be a static method just taking the Config object as a parameter and returning a String[]

cziegeler:
@cristic83 I think we were a little bit unfair here with all our comments and nit picking. I think you did very good work so far. In order to spare you the burden of applying all our nitpicking, I put your code changes into a new branch and applied our comments there.
Created a new PR based on that branch https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/97

cziegeler:
This has been implemented a while ago by #97

PR #91, opened 2023-01-17
SLING-11756: resource resolver: rewrite getVanityPathDefinition for more clarify

reschke:


anchela:
@reschke , it seems that the new ifs from the refactoring are not covered by tests. is this right?

reschke:
Yep.

reschke:
do not merge yet, I'm still testing and will move test case changes somewhere else

PR #92, opened 2023-01-17
SLING-11755: resource resolver: improve test coverage for vanity path targets

reschke:


anchela (approved):
@reschke , the tests look reasonable to me. i have to admit though that i not familiar enough with vanity paths to properly assess if something might be missing or odd.

PR #96, opened 2023-05-05
SLING-11861 - ResourceMapper.getAllMappings returns empty mapping  wh…

sagarmiglani:
…en a vanity paths is pathless url

### DIFF
--- a/src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
+++ b/src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
@@ -397,8 +397,8 @@ public class ResourceMapperImplTest {
         ExpectedMappings.existingResource("/vain-url-nopath")
             .singleMapping("/vain-url-nopath")
             .singleMappingWithRequest("/app/vain-url-nopath")
-            .allMappings("/vain-url-nopath", "")
-            .allMappingsWithRequest("/app/vain-url-nopath", "")
+            .allMappings("/vain-url-nopath", "/")
+            .allMappingsWithRequest("/app/vain-url-nopath", "/app/")
             .verify(resolver, req);
     }
 
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -1540,7 +1540,7 @@ public class MapEntries implements
             try {
                 final URL u = new URL(info);
                 prefix = u.getProtocol() + '/' + u.getHost() + '.' + u.getPort();
-                path = u.getPath();
+                path = u.getPath().isEmpty() ? "/" : u.getPath();
             } catch (final MalformedURLException e) {
                 log.warn("Ignoring malformed vanity path '{}' on {}", info, sourcePath);
                 return null;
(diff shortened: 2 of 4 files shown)


src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
```diff
@@ -397,8 +397,8 @@ public void mapResourceWithVanityPathsURLTargetNoPath() {
         ExpectedMappings.existingResource("/vain-url-nopath")
             .singleMapping("/vain-url-nopath")
             .singleMappingWithRequest("/app/vain-url-nopath")
-            .allMappings("/vain-url-nopath", "")
-            .allMappingsWithRequest("/app/vain-url-nopath", "")
```
reschke, line 401: This changes existing behavior. I believe this is very risky.

reschke (changes requested):

src/test/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImplTest.java
sagarmiglani: I agree, this is changing the behaviour, but what would be the blast radius? Could there be too many people using pathless url based vanity paths?

Also test you pointed is from "improved test coverage" \[0] (changes under SLING-11757):

```
 @Test
    public void mapResourceWithVanityPathsURLTargetNoPath() {
        ExpectedMappings.existingResource("/vain-url-nopath")
            .singleMapping("/vain-url-nopath")
            .singleMappingWithRequest("/app/vain-url-nopath")
            .allMappings("/vain-url-nopath", "")
            .allMappingsWithRequest("/app/vain-url-nopath", "")
            .verify(resolver, req);
    }
```

and it has the following expectation which itself seems wrong:

`allMappingsWithRequest("/app/vain-url-nopath", "")`


Does it look right to you? Shouldn't the there be "/app" ("" appended with the request context path) instead of ""?

When I run this test`[main] WARN org.apache.sling.resourceresolver.impl.mapping.ResourceMapperImpl - map: Unable to mangle namespaces for  returning unmangled`

[0]: https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/94/files#diff-f6a4a27e96e7b9934452dc38379974ed9583083294ebe795081b1963ac4b2c6fR401

reschke: When I added the tests, the goal was to capture the current functionality, not to reason what the right functionality is. So what we add was adding test coverage, and then fixing the NPE. That's it.

I understand you believe the behavior is wrong. You are likely right. But I have zero insight what could break when we change it. IMHO it's safer, not necessarily "better", to work around this in the calling code.

sagarmiglani: > I understand you believe the behavior is wrong. You are likely right. But I have zero insight what could break when we change it. IMHO it's safer, not necessarily "better", to work around this in the calling code.

I agree with you, but atleast we should improve the API documentation that `getAllMappings` could return "" as well in some cases. (i personally feel blast radius of this should be small and we should correct the behaviour here)

cc: @cziegeler @rombert

cziegeler: I would prefer not changing the behaviour and maybe just update the javadocs

PR #97, opened 2023-06-03
Issues/sling 11742

cziegeler:
This is an updated version of #89 , incorporating all the suggestions we did there

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
@@ -81,46 +81,6 @@ public class ResourceResolverFactoryActivator {
         public volatile CommonResourceResolverFactoryImpl commonFactory;
     }
 
-    private static final class VanityPathConfigurer {
-        private final Logger logger = LoggerFactory.getLogger(this.getClass());
-
-        void configureVanityPathPrefixes(String[] pathPrefixes, String[] pathPrefixesFallback,
-                                         String pathPrefixesPropertyName, String pathPrefixesFallbackPropertyName,
-                                         Consumer<String[]> filteredPathPrefixesConsumer) {
-            if (pathPrefixes != null) {
-                configureVanityPathPrefixes(pathPrefixes, filteredPathPrefixesConsumer);
-            } else {
-                logger.debug("The " + pathPrefixesPropertyName + " was null. Using the " +
-                    pathPrefixesFallbackPropertyName + " instead if defined.");
-                if (pathPrefixesFallback != null) {
-                    configureVanityPathPrefixes(pathPrefixesFallback, filteredPathPrefixesConsumer);
-                }
-            }
-        }
-
-        private static void configureVanityPathPrefixes(String[] pathPrefixes, Consumer<String[]> pathPrefixesConsumer) {
-            final List<String> filterVanityPaths = filterVanityPathPrefixes(pathPrefixes);
-            if (filterVanityPaths.size() > 0) {
-                pathPrefixesConsumer.accept(filterVanityPaths.toArray(new String[filterVanityPaths.size()]));
-            }
-        }
-
-        @NotNull
-        private static List<String> filterVanityPathPrefixes(String[] vanityPathPrefixes) {
-            final List<String> prefixList = new ArrayList<>();
-            for (final String value : vanityPathPrefixes) {
-                if (value.trim().length() > 0) {
-                    if (value.trim().endsWith("/")) {
-                        prefixList.add(value.trim());
-                    } else {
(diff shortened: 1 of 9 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/VanityPathConfigurer.java
```diff
@@ -0,0 +1,142 @@
...
+    private boolean isDefined(final String[] value) {
+        return value != null && value.length > 0;
+    }
+
+    List<String> configureVanityPathPrefixes(final String[] pathPrefixes, final String[] pathPrefixesFallback,
+                                     String pathPrefixesPropertyName, String pathPrefixesFallbackPropertyName) {
+        if (isDefined(pathPrefixes) && isDefined(pathPrefixesFallback) ) {
+            logger.error("Both properties, " + pathPrefixesPropertyName + " and " + pathPrefixesFallbackPropertyName
+                + ", were defined. Using " + pathPrefixesPropertyName + " for configuring vanity paths. "
+                + "Please remove the other property from your configuration.");
```
jsedding, line 117:
```suggestion
            logger.error("Both properties, \"{}\" and the deprecated \"{}\""
                + ", were defined. Using \"{}\" for configuring vanity paths. "
                + "Please remove the other property from your configuration.", pathPrefixesPropertyName, pathPrefixesFallbackPropertyName, pathPrefixesPropertyName);
```

src/main/java/org/apache/sling/resourceresolver/impl/VanityPathConfigurer.java
```diff
@@ -0,0 +1,142 @@
...
+            return filterVanityPathPrefixes(pathPrefixes);
+        } else if (isDefined(pathPrefixes)) {
+            return filterVanityPathPrefixes(pathPrefixes);
+        } else if (isDefined(pathPrefixesFallback)) {
+            logger.warn("The property " + pathPrefixesPropertyName + " was not set. Using the " +
+                pathPrefixesFallbackPropertyName + " instead. Please update your configuration to use " + pathPrefixesPropertyName);
```
jsedding, line 123:
```suggestion
            logger.warn("The property \"{}\" was not set. Using the deprecated property \"{}\""
                 + " instead. Please update your configuration to use \"{}\".", pathPrefixesPropertyName, pathPrefixesFallbackPropertyName, pathPrefixesPropertyName);
```

src/main/java/org/apache/sling/resourceresolver/impl/VanityPathConfigurer.java
```diff
@@ -0,0 +1,142 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import org.apache.sling.resourceresolver.impl.mapping.MapConfigurationProvider.VanityPathConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class VanityPathConfigurer {
```
jsedding, line 29: Why not `VanityPathConfiguration` or `AggregateVanityPathConfiguration`? `Configurer` sounds like it sets the configuration (at least it does to me), which it doesn't. It only provides an aggregated view of the configuration that is already set.

src/main/java/org/apache/sling/resourceresolver/impl/VanityPathConfigurer.java
```diff
@@ -0,0 +1,142 @@
...
+        for (final String value : vanityPathPrefixes) {
+            if (value.trim().length() > 0) {
+                if (value.trim().endsWith("/")) {
+                    prefixList.add(value.trim());
+                } else {
+                    prefixList.add(value.trim().concat("/"));
+                }
+            }
+        }
+        return prefixList.isEmpty() ? null : prefixList;
```
jsedding, line 140: I bet this could look nicer with `Stream`s and `Optional`! But, of course, that's a matter of taste.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -145,6 +145,11 @@ private static final class FactoryRegistration {
     /** Factory registration. */
     private volatile FactoryRegistration factoryRegistration;
 
+    private final VanityPathConfigurer vanityPathConfigurer = new VanityPathConfigurer();
```
jsedding, line 148: Why don't we make this a volatile field and `VanityPathConfigurer` (or preferably  `VanityPathConfiguration`) immutable?

jsedding (approved):
LGTM. I made some minor suggestions, but I see nothing fundamentally wrong with your PR. Also, I prefer this approach, where we're not updating/changing configurations dynamically.

PR #100, opened 2023-09-06
SLING-12019 - Avoid duplicate ResourceResolverFactory registrations

jsedding:


jsedding:
@cziegeler here's a proposal for a fix. I would appreciate your thoughts.

src/main/java/org/apache/sling/resourceresolver/impl/providers/stateful/AuthenticatedResourceProvider.java
```diff
@@ -52,7 +52,7 @@ public class AuthenticatedResourceProvider {
 
     private static final Logger logger = LoggerFactory.getLogger(ResourceResolverImpl.class);
 
-    public static final AuthenticatedResourceProvider UNAUTHENTICATED_PROVIDER = new AuthenticatedResourceProvider(null, false, null, null);
```
jsedding, line 55: My IDE complained and failed the tests because `null` is being passed into `@NotNull` constructor arguments. That's why I fixed this.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -490,69 +555,26 @@ protected void deactivate() {
...
-                        // nothing to do
-                    }
-                }, serviceProps);
-
-            local.runtimeRegistration = localContext.registerService(RuntimeService.class,
-                    this.getRuntimeService(), null);
-
-            this.factoryRegistration = local;
+        if (!this.isDeactivating) {
+            synchronized (this) {
```
jsedding, line 573: We could probably get rid of the synchronization. `factoryRegistration` should only ever get set via the `factoryRegistrationWorker` thread and is this guaranteed not to be concurrent.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryActivator.java
```diff
@@ -490,69 +555,26 @@ protected void deactivate() {
     private void unregisterFactory() {
         FactoryRegistration local = null;
         synchronized ( this ) {
```
jsedding, line 557: We could probably get rid of the synchronization. `factoryRegistration` should only ever get set via the `factoryRegistrationWorker` thread and is this guaranteed not to be concurrent.

jsedding:
The build failure is related to Java 17. I will take a look to see if I can get it running.

pom.xml
```diff
@@ -54,6 +54,7 @@
             <plugin>
                 <groupId>org.apache.sling</groupId>
                 <artifactId>sling-maven-plugin</artifactId>
+                <version>3.0.2</version>
```
kwin, line 57: this is part of latest parent, rather update parent version.

jsedding: Thanks @kwin ! It looks like Jörg already integrated the updates for java 17 into master, so now I merged them in instead of this commit.

src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Override
+    public void close() {
+        unregisterFactory();
+        factoryRegistrationWorker.shutdown();
+        try {
+            if (!factoryRegistrationWorker.awaitTermination(5, TimeUnit.SECONDS)) {
+                final List<Runnable> runnables = factoryRegistrationWorker.shutdownNow();
+                if (runnables.size() >= 2) {
```
rombert, line 66: This deserves a comment/method, I can't tell why we check for more than two runnables or why we ignore the first two.

src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
```diff
@@ -0,0 +1,178 @@
...
+        LOG.debug("performing unregisterFactory, factoryRegistration == {}", factoryRegistration);
+        if (this.factoryRegistration != null) {
+            this.factoryRegistration.unregister();
+            this.factoryRegistration = null;
+        }
+        LOG.debug("finished performing unregisterFactory, factoryRegistration == {}", factoryRegistration);
+    }
+
+    private static void withThreadName(String threadName, Runnable task) {
+        final String name = Thread.currentThread().getName();
```
rombert, line 128: Are you trying to expose what is going on via the thread name? It might be better for observability to keep a stable thread name prefix and reflect the current task in the second part, e.g. `getClass().getSimpleName() + " - " threadName`

rombert (approved):
I made a very high-level review and this looks good to me. IIUC the point is to use a work queue for activation/deactivations, which should solve the problem we're seeing.

A couple of small questions/nits, but nothing blocking.

I did not review whether the existing invariants ( when to register, required providers, etc ) hold as I am not familiar with those anyway.

src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
jsedding: I found it helpful to have the registration/deregistration reflected in the thread-name, because service registrations trigger further service registrations etc. So it helps understand, which changes in the system happened due to the registration/deregistration.

Regarding the naming of the thread, yes the class-name as a prefix would work. De-facto, I am using the prefix "ResourceResolverFactory". I can incorporate that into the method, if you think it would be valuable.

src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
jsedding: I think you found an artifact from a previous iteration of the patch. The intention was to make sure that the RRF is unregistered before we exit the `close()` method. I'll fix the code accordingly.

src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
rombert: Ah, that works for me. Maybe extract the prefix as a constant value somewhere, just to make it clear?

src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
jsedding: Fixed with https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/100/commits/64a00364721b44d9319219ce9ac0e102fa8032fa

src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
jsedding: Sure 👍

jsedding: Fixed in https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/100/commits/b28d4d0b208ea29579566f0b9bd823a1a3933af3

src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
rombert: Thanks!

PR #101, opened 2023-09-09
SLING-11715 in the optimized code path avoid resource resolution

joerghoh:
When the optimizedAliasResolution is enabled, the only reason why the ``current`` resource is used is to get the name from it. But that is something we don't need the resource for, but it can be done completely using the path and some basic String operations.

To achieve this I refactored the ``readAliases`` and moved the code path for the optimized resolution into a dedicated method ``readAliasesOptimized``, which does not require any Resource object anymore.
The distinction between the optimized resolution and the "standard way" is done now in the ``resolveAliases`` method, which invokes the ``readAliases`` and the ``readAliasesOptimized`` methods, and performs the Resource operations only when using the non-optimized code path.

With this approach the optimized resolution avoids multiple Resource resolutions (a single "resolve" call for each segment of the path of the resource), which should speed up this process quite a bit.

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
@@ -203,23 +203,7 @@ public class ResourceMapperImpl implements ResourceMapper {
         // make sure to append resolutionPathInfo, if present
         pathBuilder.setResolutionPathInfo(resolutionPathInfo);
         
-        Resource current = res;
-        String path = res.getPath();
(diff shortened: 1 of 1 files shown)


joerghoh:
(for some reasons the local rebase failed and I merged another commit from the master branch in here ... had to force-push to undo that)

joerghoh:
I want to wait for the outcome of https://issues.apache.org/jira/browse/SLING-12025 before merging

joerghoh:
@reschke @cziegeler I am pretty confident here, but can you have a look?

cziegeler (approved):
looks good to me

PR #103, opened 2023-09-13
SLING-12020: resource resolver: correct name of alias metric

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
```diff
@@ -77,11 +77,11 @@ public class ResourceResolverMetrics {
     private ServiceRegistration<Gauge<Long>> numberOfVanityPathBloomFalsePositiveGauge;
     private Supplier<Long> numberOfVanityPathBloomFalsePositiveSupplier = ZERO_SUPPLIER;
 
-    // number of aliases
-    private ServiceRegistration<Gauge<Long>> numberOfAliasesGauge;
-    private Supplier<Long> numberOfAliasesSupplier = ZERO_SUPPLIER;
+    // number of resources with aliased children
+    private ServiceRegistration<Gauge<Long>> numberOfResourcesWithAliasedChildrenGauge;
+    private Supplier<Long> numberOfResourcesWithAliasedChildrenSupplier = ZERO_SUPPLIER;
```
jsedding, line 82: Not really this PR, just related: I think all `Supplier` fields need to be volatile. It looks like they can be updated at any moment. Am I missing something?

jsedding (approved):
LGTM. Just one slightly related comment/question. But that's not a blocker.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
reschke: Maybe. In practive it's only set once in MapEntries's init code (not it's a supplier, not the actual value).

PR #104, opened 2023-09-15
SLING-12019 - Avoid duplicate ResourceResolverFactory registrations

jsedding:
- reworked synchronization from relying on worker queue to locks
- increased test coverage, particularly for concurrent scenarios
- switched to junit-platform to allow mixing junit4 and junit5 tests
- enabled parallel test execution for junit-platform

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
@@ -31,9 +31,11 @@ import org.slf4j.LoggerFactory;
 import java.util.Dictionary;
 import java.util.Hashtable;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
 
 public class FactoryRegistrationHandler implements AutoCloseable {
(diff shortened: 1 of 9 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
```diff
@@ -30,45 +30,85 @@
...
+            LOG.debug("activator differs = {}, factoryPreconditions differ = {}", this.activator != activator, !Objects.equals(this.factoryPreconditions, factoryPreconditions));
+            LOG.debug("factoryPreconditions {} vs {}", this.factoryPreconditions, factoryPreconditions);
+            this.factoryPreconditions = factoryPreconditions;
+            this.activator = activator;
+        } finally {
+            configurationLock.unlock();
+        }
+
+        if (reRegister) {
+            if (this.factoryRegistration != null) {
```
rombert, line 88: Should this check be done inside the critical region protected by the `configurationLock`?

src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
```diff
@@ -77,15 +117,24 @@ public void close() {
...
+            try {
+                configurationLock.lock();
+                preconditionsOk = factoryPreconditions.checkPreconditions(unavailableName, unavailableServicePid);
+                localActivator = this.activator;
+            } finally {
+                configurationLock.unlock();
+            }
+
+            factoryRegistrationWorker.execute(() -> {
+                if (preconditionsOk && this.factoryRegistration == null) {
```
rombert, line 131: What are the consequences of making an "unsychronized" read of `this.factoryRegistration` in this method?

rombert (commented):
I haven't grokked this fully, but I least as I can ask some questions :-)

src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
jsedding: You are right to spot the missing lock. But the `registrationLock` is required when accessing `factoryRegistration`.

src/main/java/org/apache/sling/resourceresolver/impl/FactoryRegistrationHandler.java
jsedding: Again, same as abov. I'll add the `registrationLock` here.

jsedding:
Thanks @rombert! While fixing the missing locks, I realized that many of the null checks on `factoryRegistration` were unnecessary. And once I had removed them, it seemed clearer to use an `AtomicReference` for the `factoryRegistration` instead of the locks.

rombert (approved):

PR #107, opened 2023-11-17
SLING-12160 Prevent ConcurrentModificationException in

kwin:
readAliasesOptimized(...)


joerghoh (approved):
lgtm

kwin:
@joerghoh Any idea if the test failures are related?

joerghoh:
@kwin Checking the details, it seems to me that alias handling in general is not guarded against concurrent modification, so just fixing this method is not good enough (besides causing these failures, which I can reproduce locally).

rombert (commented):
Fix looks good to me. I wonder whether we should take a step back and figure out the thread-safety story rather than stopping at a minimal fix.

That can come separately from the initial fix, of course.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
```diff
@@ -291,7 +292,7 @@ private List<String> readAliasesOptimized(String path) {
     	if ( aliases == null || !aliases.containsValue(name) ) 
     		return Collections.emptyList();
 
-    	return aliases.entrySet().stream()
+    	return new LinkedHashMap<>(aliases).entrySet().stream()
```
jsedding, line 295: Doesn't the `LinkedHashMap` constructor iterate over the `aliases` map? If that's the case nothing is solved.

kwin: Indeed, I am looking into reproducing the concurrency issues in a unit test and afterwards come up with a better fix.

jsedding (commented):
@kwin since you're changing the internal datastructure, I was wondering if it wouldn't make sense to encapsulate the alias handling in its own dedicated object. That could make testing a lot easier, and it would even pave the way for experimenting with different implementations, if we were ever so inclined.

E.g. I could imagine such an implementation based on `org.apache.commons.collections4.trie.PatriciaTrie` could be interesting. Mind you, the `PatriciaTree` is not thread safe, so in our hypothetical `AliasMapper`, we would need to synchronize access to it.

Just some thoughts. Feel free to pursue the current direction, which looks ok to me. I only didn't approve (yet), because I'd like a better understanding of the questions I asked.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -140,7 +142,10 @@ public class MapEntries implements
 
     private Map <String,List <String>> vanityTargets;
 
-    private Map<String, Map<String, String>> aliasMap;
+    /**
+     * The key of the map is the parent path, while the value is a map with the the resource name as key and the actual aliases as values)
+     */
+    private ConcurrentMap<String, ConcurrentMap<String, List<String>>> aliasMapsMap;
```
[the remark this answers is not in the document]
kwin: > If I understand correctly you're changing the data structure from "Map<parentPath, Map<aliasedName, resourceName>" to "Map<parentPath, Map<resourceName, List>". Is that correct? If that is the case, I like the more compact storage. Just be aware, however, that it makes "lookup by alias" slightly more expensive, as it requires traversal of the entire map returned for a parentPath, including a Collection#contains(String) call on the List. On the flip side, it is cheaper to generate all aliased paths.

Yes, this is all correct. I consider the case to lookup an alias for a given resource name (i.e. map) more frequent than looking for an alias below a specific parent (i.e. resolve). The flipping of keys and values is necessary to keep the order of the `sling:alias` values for a specific resource (see next question)

> Rgd. the implementation, I would avoid using ConcurrentMap in the type definition (unless we use some of its API), and rather use Map<String, Map<String, List<String>>.

This was done deliberately to make users aware that this is a thread safe implementation, i.e. no need to synchronize access on it. This is more explicit than just mentioning it in the javadoc. What issues do you see we that?

> Do we need to preserve their order?

Yes, because the first entry is used for Outgoing Mapping

> Do we need to maintain duplicates?

Only to emit a warn once if there is a duplicate found in siblings. The duplicates are just discarded, this is exactly the behaviour as before

> Is the list updated or replaced? And, given the above questions, would it make sense to replace rather than update it?

The implementation used here `CopyOnWriteArrayList` uses a replacement internally. This is easier to deal with than recreating the list manually for updates. For me this impl is reasonable here, because we can only expect at most a handful of items per list and this impl. doesn't come with any performance drawback during reading, just a small one when updating (due to the copy of the list)

kwin:
> I was wondering if it wouldn't make sense to encapsulate the alias handling in its own dedicated object

In general I agree, but currently the tests are relying on implementation details so that a refactoring would require a rewrite of most of the tests. I consider this too risky for this change and would rather approach that separately (i.e. do that without doing any functional change)

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -18,32 +18,6 @@
  */
 package org.apache.sling.resourceresolver.impl.mapping;
 
-import org.apache.commons.collections4.map.LRUMap;
```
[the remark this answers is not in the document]
kwin: Agreed, done in https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/107/commits/97ca6165edfc7c913575d7ac65b10813e6c9acba.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1243,24 +1236,20 @@ private boolean loadAlias(final Resource resource, Map<String, Map<String, Strin
...
+                        Optional<String> currentResourceName = parentMap.entrySet().stream().filter(entry -> entry.getValue().contains(alias)).findFirst().map(Map.Entry::getKey);
+                        if (currentResourceName.isPresent()) {
                             log.warn(
                                     "Encountered duplicate alias {} under parent path {}. Refusing to replace current target {} with {}.",
-                                    alias, parentPath, current, resourceName);
+                                    alias, parentPath, currentResourceName.get(), resourceName);
                         } else {
-                            parentMap.put(alias, resourceName);
+                            List<String> existingAliases = parentMap.computeIfAbsent(resourceName, name -> new CopyOnWriteArrayList<>());
+                            existingAliases.add(alias);
```
jsedding, line 1252: We're not accurately mirroring the ordering of the `aliasArray` into the list. I really think it would be better to construct a `LinkedHashSet`, fill it with the aliases, wrap it using `Collections#unmodifieableSet()` and at the end replace it in the map.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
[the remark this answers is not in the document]
jsedding: > Yes, this is all correct. I consider the case to lookup an alias for a given resource name (i.e. map) more frequent than looking for an alias below a specific parent (i.e. resolve). The flipping of keys and values is necessary to keep the order of the sling:alias values for a specific resource (see next question) as there is no ConcurrentMap keeping the insertion order (similar to LinkedHashMap).

Agreed. And I see the dilemma of not having a concurrent Map implementation that maintains the insertion order. I also prefer your proposed datastructure, because the alias ordering is localized to the individual resource and thus easier to understand.

Regarding the "easier to understand" part, I believe that the current implementation, as well as yours, suffer from an underspecification of the ordering of aliases defined on a resource _and_ on its `jcr:content` child. I.e. assume we have `/foo/sling:alias=bar` and  `/foo/jcr:content/sling:alias=baz`. Which one would be "first"? IIUC, currently it depends on which one is inserted first.

> This was done deliberately to make users aware that this is a thread safe implementation, i.e. no need to synchronize access on it. This is more explicit than just mentioning it in the javadoc. What issues do you see we that? ConcurrentMap and Map share exactly the same interface signature, the differences are only in the default implementations.

I would much rather hide the implementation details, especially in the signatures. This makes future refactoring easier. That's why I would really like to see the functionality encapsulated in an object, rather than it being plastered all over the `MapEntries` class. I agree with you, however, that we would first need to re-write the tests to be agnostic of implementation details. They are, unfortunately, quite horrible, IMHO :(

> The implementation used here CopyOnWriteArrayList uses a replacement internally. [...]

I still believe we would be better of creating a new immutable list (or preferably a set) each time, and replace it. Immutable objects are always easier to reason about than mutable ones. And updating the list to reflect the correct order of aliases is harder than creating a new one IMHO.

jsedding (commented):
Given your comments, I think we should really start with re-writing the tests. At the moment I see no trivial, low risk way to fix the concurrency issue without doing that.

Sorry, I realize you invested quite some time already. But I don't feel comfortable with the changes as they are. Particularly, because the resourceresolver is used all the time.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
kwin: I think we do because we loop through the value in the order they are given in the property (https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/107/files/97ca6165edfc7c913575d7ac65b10813e6c9acba#diff-b8232d361560d98697fcb8774c43f8f4a33e213a1c892171d339f956ef73df04R1240). Why do you want to replicate the logic from `CopyOnWriteArrayList`. Important for me is that this is ordered not so much that it doesn't contain duplicates, therefore I went for List<> instead of SortedSet<>

kwin:
> Regarding the "easier to understand" part, I believe that the current implementation, as well as yours, suffer from an underspecification of the ordering of aliases defined on a resource and on its jcr:content child. I.e. assume we have /foo/sling:alias=bar and /foo/jcr:content/sling:alias=baz. Which one would be "first"? IIUC, currently it depends on which one is inserted first.

The logic has not changed so therefore I consider this argument orthogonal to the actual issue being fixed here. Feel free to create a dedicated issue to get some clarification or come up with a use case where this impl behaves different than before.

> I would much rather hide the implementation details, especially in the signatures

We can evolve after applying this fix. It is deliberately not doing refactoring to be easier to review. But let me revert to using the most generic interfaces for now which is `Map<Map<String,Collection<String>>`

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
jsedding: I think we don't because if we change a `sling:alias` property from `["foo", "bar"]` to  `["baz", "foo", "bar"]`, then "baz" would be added to the end of the list. That said, I believe the behaviour before your changes didn't account for this scenario either. Therefore, I don't consider it a blocker.

jsedding (approved):
@kwin I think your patch fixes the `ConcurrentModificationException` and otherwise it doesn't leave us in a worse place than where we were before. So I'll approve the PR.

If you could double-check rgd. unused imports before merging, that would be great!

Thanks for bearing with me, I learned a lot about the code in the process.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
```diff
@@ -161,8 +161,8 @@ public void setup() throws Exception {
 
         Optional<ResourceResolverMetrics> metrics = Optional.empty();
 
-        mapEntries = Mockito.spy(new MapEntries(resourceResolverFactory, bundleContext, eventAdmin, stringInterpolationProvider, metrics));
-        final Field aliasMapField = MapEntries.class.getDeclaredField("aliasMap");
+        mapEntries = /*Mockito.spy(*/new MapEntries(resourceResolverFactory, bundleContext, eventAdmin, stringInterpolationProvider, metrics);//);
```
kwin, line 164: @jsedding Do you have any idea why the spy was used here? I commented it to ease debugging and not having it didn't lead to any test failures.

jsedding: Nope. No idea.

PR #108, opened 2023-12-20
SLING-12204: ResourceResolver: refactor PagedQueryIterator for re-use

reschke:
This generalizes the PagedQueryIterator so it can be used for other purposes (such as sling:alias) later.

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -1303,6 +1303,7 @@ public class MapEntries implements
         private int page = 0;
         private int pageSize;
         private Resource next = null;
+        private String[] defaultValue = new String[0];
 
         public PagedQueryIterator(String name, String propertyName, ResourceResolver resolver, String query, int pageSize) {
             this.name = name;
@@ -1327,7 +1328,7 @@ public class MapEntries implements
         private Resource getNext() throws NoSuchElementException {
             Resource resource = it.next();
             count += 1;
-            final String[] values = resource.getValueMap().get(propertyName, new String[0]);
+            final String[] values = resource.getValueMap().get(propertyName, defaultValue);
             if (values.length > 0) {
                 String value = values[0];
                 if (value.compareTo(lastValue) < 0) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1288,50 +1288,60 @@ private Iterator<Resource> queryAllVanityPaths(String query) {
...
         private Iterator<Resource> it;
         private int count = 0;
         private int page = 0;
-        private int pageSize = Integer.getInteger("sling.vanityPath.pageSize", 2000);
+        private int pageSize;
         private Resource next = null;
+        private String[] defaultValue = new String[0];
 
-        public PagedQueryIterator(ResourceResolver resolver, String query) {
+        public PagedQueryIterator(String name, String propertyName, ResourceResolver resolver, String query, int pageSize) {
```
joerghoh, line 1308: ```name``` is quite generic. If we don't come up with a better name :-) we should at least document its purpose here.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1288,50 +1288,60 @@ private Iterator<Resource> queryAllVanityPaths(String query) {
         return it;
     }
 
+    /**
+     * Utility class for running paged queries.
+     */
     private class PagedQueryIterator implements Iterator<Resource> {
```
joerghoh, line 1294: would it make sense to make it a full class now (instead of an inner class)?

reschke: Maybe once we need it outside MapEntries. Wanted to miminize the diff for now.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1288,50 +1288,66 @@ private Iterator<Resource> queryAllVanityPaths(String query) {
...
         private int count = 0;
         private int page = 0;
-        private int pageSize = Integer.getInteger("sling.vanityPath.pageSize", 2000);
+        private int pageSize;
         private Resource next = null;
+        private String[] defaultValue = new String[0];
 
-        public PagedQueryIterator(ResourceResolver resolver, String query) {
+        /**
+         * @param subject name of the query
```
joerghoh, line 1309:
```suggestion
         * @param subject name of the query, will be used only for logging
```

joerghoh (approved):

PR #109, opened 2023-12-20
SLING-12205: ResourceResolver: switch sling:alias query to JCR-SQL2

reschke:
(needed for subsequent use of query paging)

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -112,8 +112,6 @@ public class MapEntries implements
 
     private static final String JCR_SYSTEM_PREFIX = JCR_SYSTEM_PATH + '/';
 
-    static final String ALIAS_BASE_QUERY_DEFAULT = "SELECT sling:alias FROM nt:base AS page";
-
     static final String ANY_SCHEME_HOST = "[^/]+/[^/]+";
 
     /** default log */
@@ -1154,7 +1152,7 @@ public class MapEntries implements
 
         log.debug("start alias query: {}", queryString);
         long queryStart = System.nanoTime();
-        final Iterator<Resource> i = resolver.findResources(queryString, "sql");
+        final Iterator<Resource> i = resolver.findResources(queryString, "JCR-SQL2");
         long queryElapsed = System.nanoTime() - queryStart;
         log.debug("end alias query; elapsed {}ms", TimeUnit.NANOSECONDS.toMillis(queryElapsed));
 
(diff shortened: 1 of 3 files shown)


joerghoh (approved):

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1154,7 +1152,7 @@ private Map<String, Map<String, Collection<String>>> loadAliases(final ResourceR
 
         log.debug("start alias query: {}", queryString);
         long queryStart = System.nanoTime();
-        final Iterator<Resource> i = resolver.findResources(queryString, "sql");
+        final Iterator<Resource> i = resolver.findResources(queryString, "JCR-SQL2");
```
enapps-enorman, line 1155: I'd suggest using the Query.JCR_SQL2 constant here instead.

[Query JavaDocs](https://javadoc.io/static/javax.jcr/jcr/2.0/javax/jcr/query/Query.html)

src/test/java/org/apache/sling/resourceresolver/impl/mapping/InMemoryResourceProvider.java
```diff
@@ -105,7 +105,7 @@ public Iterator<Resource> findResources(@NotNull ResolveContext<Void> ctx, Strin
 
                 // we don't explicitly filter paths under jcr:system, but we don't expect to have such resources either
                 // and this stub provider is not the proper location to test JCR queries
-                if  ( "sql".equals(language) && "SELECT sling:alias FROM nt:base AS page WHERE (NOT ISDESCENDANTNODE(page,'/jcr:system')) AND sling:alias IS NOT NULL".equals(query) ) {
+                if  ( "JCR-SQL2".equals(language) && "SELECT [sling:alias] FROM [nt:base] WHERE NOT isdescendantnode('/jcr:system') AND [sling:alias] IS NOT NULL".equals(query) ) {
```
enapps-enorman, line 108: I'd suggest using the Query.JCR_SQL2 constant here.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
```diff
@@ -187,7 +187,7 @@ public void test_simple_alias_support() {
         when(result.getName()).thenReturn("child");
         when(result.getValueMap()).thenReturn(buildValueMap(ResourceResolverImpl.PROP_ALIAS, "alias"));
 
-        when(resourceResolver.findResources(anyString(), eq("sql"))).thenAnswer(new Answer<Iterator<Resource>>() {
+        when(resourceResolver.findResources(anyString(), eq("JCR-SQL2"))).thenAnswer(new Answer<Iterator<Resource>>() {
```
enapps-enorman, line 190: I'd suggest using the Query.JCR_SQL2 constant here.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
```diff
@@ -224,7 +224,7 @@ public void test_that_duplicate_alias_doesnt_replace_first_alias() {
         when(secondResult.getName()).thenReturn("child2");
         when(secondResult.getValueMap()).thenReturn(buildValueMap(ResourceResolverImpl.PROP_ALIAS, "alias"));
 
-        when(resourceResolver.findResources(anyString(), eq("sql"))).thenAnswer(new Answer<Iterator<Resource>>() {
+        when(resourceResolver.findResources(anyString(), eq("JCR-SQL2"))).thenAnswer(new Answer<Iterator<Resource>>() {
```
enapps-enorman, line 227: I'd suggest using the Query.JCR_SQL2 constant here.

enapps-enorman (commented):
Just a couple of suggestions about using the existing query language constants.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: Not convinced. We currently do not use javax.jcr imports; likely because Sling in theory supports other backends.

I personally do not buy this argument, but if we go down that road, there's a lot more we could / should change.

enapps-enorman: Pretty weak that you just merging immediately without giving me a chance to respond.   You are already using a JCR specific language there so pretending that a different backend would support that language somehow seems unlikely to me.  Using a string doesn't magically make it more portable.

rombert: We already optionally import `javax.jcr`, so I don't see what the issue would be with referencing the constant value.

https://github.com/apache/sling-org-apache-sling-resourceresolver/blob/c7bdc4d6a5e360623b0ad3c117f1728e26c12e8e/bnd.bnd#L2

The java compiler will inline string references in the resulting class file anyway.

enapps-enorman: > The java compiler will inline string references in the resulting class file anyway.

Yes, this is true.  I prefer using the Query.JCR_SQL2 constant just in case that language gets deprecated someday (like what happened for the "xpath" and "sql" languages) so that your IDE will warn you about usage of deprecated code in that scenario.

reschke:
FWIW, I chose not to act on the feedback above because it's not specific to *this* PR. I'm ok with having a discussion unsing javax.jcr in the source, but we haven't before, and that change should me made consistently, not just in that part of the code.

enapps-enorman:
> it's not specific to _this_ PR

@reschke  I'm not sure that is true.  You specifically changed the code to use a new query language.  For me, using the best practices for that new query language is the appropriate thing to do.  If you do it correctly the the first time then you don't have to waste time re-doing it later.

PR #111, opened 2023-12-21
SLING-11799: ResourceResolver - use keyset pagination for sling:alias query

reschke:
Changes query strategy for sling:alias to the same as sling:vanityPath.

Fallbacks allow deploying on older store versions or on stores that require an index on "first(sling:alias)".

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -112,8 +112,6 @@ public class MapEntries implements
 
     private static final String JCR_SYSTEM_PREFIX = JCR_SYSTEM_PATH + '/';
 
-    static final String ALIAS_BASE_QUERY_DEFAULT = "SELECT sling:alias FROM nt:base AS page";
-
     static final String ANY_SCHEME_HOST = "[^/]+/[^/]+";
 
     /** default log */
@@ -1154,7 +1152,7 @@ public class MapEntries implements
 
         log.debug("start alias query: {}", queryString);
         long queryStart = System.nanoTime();
-        final Iterator<Resource> i = resolver.findResources(queryString, "sql");
+        final Iterator<Resource> i = resolver.findResources(queryString, "JCR-SQL2");
         long queryElapsed = System.nanoTime() - queryStart;
         log.debug("end alias query; elapsed {}ms", TimeUnit.NANOSECONDS.toMillis(queryElapsed));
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1148,19 +1148,26 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
-        log.debug("start alias query: {}", queryString);
-        long queryStart = System.nanoTime();
-        final Iterator<Resource> i = resolver.findResources(queryString, "JCR-SQL2");
-        long queryElapsed = System.nanoTime() - queryStart;
-        log.debug("end alias query; elapsed {}ms", TimeUnit.NANOSECONDS.toMillis(queryElapsed));
+        Iterator<Resource> it;
+        try {
+            final String queryStringWithSort = baseQueryString + " AND FIRST([sling:alias]) > '%s' ORDER BY FIRST([sling:alias])";
+            it = new PagedQueryIterator("alias", "sling:alias", resolver, queryStringWithSort,
+                    Integer.getInteger("sling.alias.pageSize", 2000));
```
rombert, line 1157: This should come as an OSGi config, not as a system property. OSGi configuration properties can then be picked up from environment variables if desired, see https://github.com/apache/felix-dev/blob/master/configadmin-plugins/interpolation/README.md#interpolating-environment-variables .

The OSGi config should have that default of 2000 set.

reschke: The system property was added to ease debugging and performance testing, not to be set in practice.

I seriously doubt that adding more OSGi toggles would be helpful, but if we did this, we should do it consistently (and maybe dissuss whether alias and vanity path need different ones).

rombert: When we should remove this, right?

reschke: Not necessarily; more debugging and performance testing may be needed.

In any case: this is consistent with what we do for vanity paths.

reschke: @joerghoh and @cziegeler - what do you think?

joerghoh: @reschke IIUC this is an implementation detail, which is not impacting the functionality in any way. It could influence performance under some circumstances though.

I don't think that it makes sense to expose this setting via OSGI configuration, as it would indicate that there might be reasons to change it. Not sure if even the chance to override it via system properties is necessary.

Can we postpone this decision until we have gathered some experience with it? Until then we should treat the way to change the value via system properties as implementation detail, which can change at any time.

joerghoh (approved):
lgtm

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
bdelacretaz: If there's agreement on using the system property for experimenting only, I think defining it as below would clarify that:

    // Using a system property here, as changing this value only makes sense
    // for debugging and code performance testing
    private static int ALIAS_PAGE_SIZE = Integer.getInteger("sling.alias.pageSize", 2000));

rombert: Looking at the code is not enough to understand that this is intended for internal experiments only, which explains the "big" conversation I started because of it. I like @bdelacretaz's suggestion on how to extract this, and would add the suggestion to name the property "sling.unsupported.alias.pageSize", for clarity.

reschke:
FWIW, the discussion about whether or not to use the system property in is ... interesting, but blocks progress here. IMHO we should do that as part of a separate ticket, which would address the same question for sling:vanityPath as well. For sure we want consistency here, no?

bdelacretaz:
> ...the discussion about whether or not to use the system property in is ... interesting, but blocks progress here...

I don't think there are any objections left if that usage is clearly expressed in code. Combining my previous suggestion with the additional comments about it, I think defining the page size as below, instead of inline in the code, would work for all involved in this discussion:

    // Using a system property here, as changing this value only makes sense
    // for debugging and code performance testing
    private static int ALIAS_PAGE_SIZE = Integer.getInteger("sling.debug.alias.pageSize", 2000));

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: -> https://issues.apache.org/jira/browse/SLING-12225

PR #113, opened 2024-01-10
SLING-12225: remove system properties for setting query page size

reschke:


rombert (approved):
Looks good to me, thanks. It would look cleaner if the page size would be moved to constants (one for aliases, one for vanity paths) but that's not a review blocker.

PR #115, opened 2024-02-05
SLING-12244 cache the result of the getParentResourceType call

joerghoh:


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1054,14 +1058,41 @@ public boolean hasChildren(Resource resource) {
...
+     */
+    private @NotNull Map<String,String> getParentResourceTypeMap() {
+        @SuppressWarnings("unchecked")
+        Map<String,String> result = (Map<String, String>) getPropertyMap().get(PARENT_RT_CACHEKEY);
+        if (result == null) {
+            result = new HashMap<>();
+            getPropertyMap().put(PARENT_RT_CACHEKEY, result);
+        }
+        return result;
+    }
```
jsedding, line 1092: This can be simplified:

```suggestion
    private @NotNull Map<String,String> getParentResourceTypeMap() {
        @SuppressWarnings("unchecked")
        return (Map<String, String>) getPropertyMap().computeIfAbsent(PARENT_RT_CACHEKEY, k -> new HashMap<>());
    }
```

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -1054,14 +1058,41 @@ public boolean hasChildren(Resource resource) {
...
         String resourceSuperType = null;
         if ( resource != null ) {
-            resourceSuperType = resource.getResourceSuperType();
-            if (resourceSuperType == null) {
-                resourceSuperType = this.getParentResourceType(resource.getResourceType());
+            if (getParentResourceTypeMap().containsKey(resource.getPath())) {
+                resourceSuperType = getParentResourceTypeMap().get(resource.getPath());
+            } else {
+                resourceSuperType = getParentResourceTypeInternal(resource);
+                getParentResourceTypeMap().put(resource.getPath(), resourceSuperType);
```
jsedding, line 1065: Can be simplified:

```suggestion
            resourceSuperType = getParentResourceTypeMap().computeIfAbsent(resource.getPath(), p -> getParentResourceTypeInternal(resource));
```

jsedding (approved):

jsedding:
Looks good! I suggested some simplifications. Up to you if you want to apply them.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
joerghoh: I don't think that this will work, because it lead to a recursive call:
``getPrentResourceType`` -> ``getParentResourceTypeInternal`` -> ``getParentResourceType``

and that as part of ``computeIfAbsent``  multiple keys are modified in the map returned by ``getParentResourceTypeMap()``, which can lead to a ``ConcurrentModificationException``.

PR #118, opened 2024-07-16
SLING-12383: ResourceResolver: potential NPE while processing aliases

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1224,10 +1224,15 @@ private boolean loadAlias(final Resource resource, Map<String, Map<String, Colle
             containingResource = resource;
         }
 
+        if (containingResource == null) {
```
joerghoh, line 1227: I would move this check a bit up into the block after line 1222:

```
if (JCR_CONTENT.equals(resource.getName())) {
  containingResource = resource.getParent();
  if (containingResource == null) {
    log.warn(...)
    return false;
  }
} else ...
```

because only there ``containingResource`` can be assigned that null value.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1224,10 +1224,15 @@ private boolean loadAlias(final Resource resource, Map<String, Map<String, Colle
...
+            log.warn("containingResource is null for alias on {}, skipping.", resource.getPath());
+            return false;
+        }
+ 
         final Resource parent = containingResource.getParent();
 
         if (parent == null) {
-            log.debug("parent is null for alias on {}.", resource.getName());
+            log.warn("parent is null for alias on {}, skipping.", resource.getPath());
```
joerghoh, line 1235: The message is not entirely accurate, as ``parent`` could be also the grand-parent of ``resource`` :-)

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: Sounds right.

rombert (commented):
LGTM, Joerg's comments make sense.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: you mean exactly *which*? I'd pefer to do that separately as part of a separate ticket/PR.

rombert (commented):
A final nit :-) Feel free to merge without if you think it is not useful.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1220,14 +1220,18 @@ private boolean loadAlias(final Resource resource, Map<String, Map<String, Colle
...
+            }
         } else {
             containingResource = resource;
         }
 
         final Resource parent = containingResource.getParent();
 
         if (parent == null) {
-            log.debug("parent is null for alias on {}.", resource.getName());
+            log.warn("containing ancestor is null for alias on {}, skipping.", resource.getPath());
```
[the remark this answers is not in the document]
reschke: you mean in case containingResource != resource?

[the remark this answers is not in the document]
reschke: ok, I can split this into two different messages; does that make sense?

rombert (approved):

PR #119, opened 2024-07-17
SLING-12384: ResourceResolver: paged query for sling:alias will miss entries when 'first(sling:alias)' is empty

reschke:
Changes the initial query to use `>=` instead of `>` so the empty string is matched as well. Also fixes paging when key change happens on page end. Add WARN when page size appears to be bigger that it should be. Un-ignores previously failing tests.

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -1393,9 +1393,7 @@ public class MapEntries implements
 
                 // start next page?
                 if (count > pageSize && !value.equals(lastValue)) {
-                    largestPage = Math.max(largestPage, count - 1);
-                    log.debug("read {} query (page {}); {} entries, last key was: {}, largest page so far: {}", subject, page - 1,
-                            count, value, largestPage);
+                    doPageStats();
                     lastKey = value;
                     nextPage();
                     return getNext();
@@ -1412,6 +1410,7 @@ public class MapEntries implements
                 try {
                     next = getNext();
                 } catch (NoSuchElementException ex) {
+                    doPageStats();
                     // there are no more
                     next = null;
                 }
@@ -1424,6 +1423,12 @@ public class MapEntries implements
             Resource result = next != null ? next : getNext();
             next = null;
             return result;
+        }
+
+        private void doPageStats() {
+            largestPage = Math.max(largestPage, count - 1);
+            log.debug("read {} query (page {}); {} entries, last key was: {}, largest page so far: {}", subject, page - 1,
+                    count, lastKey, largestPage);
         }
 
         public String getStatistics() {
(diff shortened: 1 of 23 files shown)


rombert:
I am not well-suited to review this, JCR queries are not my strong point.

joerghoh (changes requested):

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1389,6 +1419,16 @@ public Resource next() throws NoSuchElementException {
             next = null;
             return result;
         }
+
+        public String getStatistics() {
+            String result = String.format(" (max. page size: %d, number of pages: %d)", largestPage, page);
+            if (largestKeyCount > pageSize * 10) {
+                result += String.format(" - WARNING: largest number of aliases with the same 'first' selector exceeds expectations (value '%s' appears %d times)",
```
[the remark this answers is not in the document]
reschke: Log the WARN from inside the PagedQueryIterator, or add a new getter for warnings?

joerghoh (approved):

mbaedke (approved):

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1393,6 +1437,27 @@ public Resource next() throws NoSuchElementException {
...
+        public @NotNull String getStatistics() {
+            return String.format(" (max. page size: %d, number of pages: %d)", largestPage, page);
+        }
+
+        public @NotNull String getWarning() {
+            int warnAt = pageSize * 10;
+            if (largestKeyCount > warnAt) {
+                return String.format(
+                        "Largest number of aliases with the same 'first' selector exceeds expectations (value '%s' appears %d times)",
+                        largestKeyValue, largestKeyCount, warnAt);
```
joerghoh, line 1456: I don't find a format specifier for ``warnAt``

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1393,6 +1437,27 @@ public Resource next() throws NoSuchElementException {
             next = null;
             return result;
         }
+
+        private void doPageStats() {
```
joerghoh, line 1441: The method name is a bit unspecific, maybe this?
```suggestion
        private void updatePageStats() {
```

src/test/java/org/apache/sling/resourceresolver/impl/mapping/PagedQueryIteratorTest.java
```diff
@@ -106,23 +105,39 @@ public void testSimpleWrongOrder() {
...
-            assertEquals(key, getFirstValueOf(it.next(), PROPNAME));
-        }
-        assertFalse(it.hasNext());
+        checkResult(it, expected);
+        assertEquals("", it.getWarning());
+    }
+
+    @Test
+    public void testPagedLargePage() {
+        String[] expected = new String[] { "a", "a", "a", "a", "a", "a", "a", "a", "a", "a", "a", "a", "a", "a", "a", "a", "a", "a",
```
joerghoh, line 123:
```suggestion
        String[] expected = new String[140];
        Arrays.fill(expected,"a");
```
(I haven't counted the instances of 'a', but I assume it's 140 because that number is mentioned below.)

joerghoh (changes requested):

joerghoh (approved):
lgtm

PR #122, opened 2024-08-21
SLING-12412: ResourceResolver: add test coverage for empty aliases, blank aliases, and invalid aliases

reschke:
This just increases test coverage, with no assumptions whether the observed behavior is actually correct.


joerghoh (approved):
+1 to document the existing behavior.

PR #125, opened 2024-09-13
SLING-12402: ResourceResolver: add additional log entry when there are many conflicting or invalid aliases (also add metrics)

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
```diff
@@ -85,18 +87,27 @@ public class ResourceResolverMetrics {
...
-        numberOfVanityPathBloomFalsePositiveGauge = registerGauge(bundleContext, METRICS_PREFIX + ".numberOfVanityPathBloomFalsePositives", () -> numberOfVanityPathBloomFalsePositiveSupplier );
+        numberOfVanityPathsGauge = registerGauge(bundleContext, METRICS_PREFIX + ".numberOfVanityPaths", () -> numberOfVanityPathsSupplier);
+        numberOfResourcesWithVanityPathsOnStartupGauge = registerGauge(bundleContext, METRICS_PREFIX + ".numberOfResourcesWithVanityPathsOnStartup", () -> numberOfResourcesWithVanityPathsOnStartupSupplier);
+        numberOfVanityPathLookupsGauge = registerGauge(bundleContext, METRICS_PREFIX + ".numberOfVanityPathLookups", () -> numberOfVanityPathLookupsSupplier);
+        numberOfVanityPathBloomNegativeGauge = registerGauge(bundleContext, METRICS_PREFIX + ".numberOfVanityPathBloomNegative", () -> numberOfVanityPathBloomNegativeSupplier);
+        numberOfVanityPathBloomFalsePositiveGauge = registerGauge(bundleContext, METRICS_PREFIX + ".numberOfVanityPathBloomFalsePositive", () -> numberOfVanityPathBloomFalsePositiveSupplier);
```
joerghoh, line 106: did you just remove the trailing ``s`` character from the name of the already metric? While technically an API breaking change, this could break anyone already using this metric.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -242,7 +260,16 @@ protected boolean doInit() {
...
                     final Map<String, Map<String, Collection<String>>> loadedMap = this.loadAliases(resolver);
                     this.aliasMapsMap = loadedMap;
-    
+
+                    // warn if there are more than a few defunct aliases
+                    if (conflictingAliases.size() >= MAX_REPORT_DEFUNCT_ALIASES) {
+                        log.warn("There are {} conflicting aliases; excerpt: {}", conflictingAliases.size(), conflictingAliases);
+                    }
+                    if (invalidAliases.size() >= MAX_REPORT_DEFUNCT_ALIASES) {
+                        log.warn("There are {} invalid aliases; excerpt: {}", invalidAliases.size(), invalidAliases);
```
joerghoh, line 269:
```suggestion
                        log.warn("There are more than {} invalid aliases; excerpt: {}", invalidAliases.size(), invalidAliases);
```

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -242,7 +260,16 @@ protected boolean doInit() {
                 try {
                     final Map<String, Map<String, Collection<String>>> loadedMap = this.loadAliases(resolver);
                     this.aliasMapsMap = loadedMap;
-    
+
+                    // warn if there are more than a few defunct aliases
+                    if (conflictingAliases.size() >= MAX_REPORT_DEFUNCT_ALIASES) {
+                        log.warn("There are {} conflicting aliases; excerpt: {}", conflictingAliases.size(), conflictingAliases);
```
joerghoh, line 266:
```suggestion
                        log.warn("There are more than {} conflicting aliases; excerpt: {}", conflictingAliases.size(), conflictingAliases);
```

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -242,7 +260,16 @@ protected boolean doInit() {
...
+                    }
+                    if (invalidAliases.size() >= MAX_REPORT_DEFUNCT_ALIASES) {
+                        log.warn("There are {} invalid aliases; excerpt: {}", invalidAliases.size(), invalidAliases);
+                    }
+                    conflictingAliases.clear();
```
joerghoh, line 271: why do you clear these structures? From what I understood, this code is executed only once for the lifetime of the ResourceResolverFactory, so it's not necessary at all. Nothing against it (easy saved bytes), but a comment would explain why this is done here.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1266,17 +1293,27 @@ private boolean loadAlias(final Resource resource, Map<String, Map<String, Colle
...
                         Optional<String> siblingResourceNameWithDuplicateAlias = parentMap.entrySet().stream()
                                 .filter(entry -> !entry.getKey().equals(resourceName)) // ignore entry for the current resource
                                 .filter(entry -> entry.getValue().contains(alias))
                                 .findFirst().map(Map.Entry::getKey);
                         if (siblingResourceNameWithDuplicateAlias.isPresent()) {
+                            long conflicting = detectedConflictingAliases.incrementAndGet();
                             log.warn(
-                                    "Encountered duplicate alias '{}' under parent path '{}'. Refusing to replace current target {} with {}.",
-                                    alias, parentPath, siblingResourceNameWithDuplicateAlias.get(), resourceName);
+                                    "Encountered duplicate alias '{}' under parent path '{}'. Refusing to replace current target '{}' with '{}' (total so far: {}).",
```
joerghoh, line 1311:
```suggestion
                                    "Encountered duplicate alias '{}' under parent path '{}'. Refusing to replace current target '{}' with '{}' (total duplicated aliases so far: {}).",
```

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -242,7 +260,16 @@ protected boolean doInit() {
                 try {
                     final Map<String, Map<String, Collection<String>>> loadedMap = this.loadAliases(resolver);
                     this.aliasMapsMap = loadedMap;
-    
+
+                    // warn if there are more than a few defunct aliases
+                    if (conflictingAliases.size() >= MAX_REPORT_DEFUNCT_ALIASES) {
```
joerghoh, line 265: I see it right below, elements are only added to conflictingAliases while ``conflicting < MAX_REPORT_DEFUNCT_ALIASES``. that means ``conflictingAliases.size()`` is always less than ``MAX_REPORT_DEFUNCT_ALIASES``. 


I would always log conflicts, even if we have only 5.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -242,7 +260,16 @@ protected boolean doInit() {
...
+                        log.warn("There are {} conflicting aliases; excerpt: {}", conflictingAliases.size(), conflictingAliases);
+                    }
+                    if (invalidAliases.size() >= MAX_REPORT_DEFUNCT_ALIASES) {
```
joerghoh, line 268: same as above.

joerghoh (changes requested):
some suggestions for minor log improvements.

But I have questions around the logging of the conflicting  aliases/vanityPaths, and the renaming of existing metrics.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
reschke: There was an inconstency between metric names and method names (which I noticed when adding tests). It might be better to adjust the method names, I guess.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: I see,

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -242,7 +260,21 @@ protected boolean doInit() {
                 try {
                     final Map<String, Map<String, Collection<String>>> loadedMap = this.loadAliases(resolver);
                     this.aliasMapsMap = loadedMap;
-    
+
+                    // warn if there are more than a few defunct aliases
+                    if (conflictingAliases.size() >= MAX_REPORT_DEFUNCT_ALIASES) {
+                        log.warn("There are {} {} conflicting aliases; excerpt: {}",  conflictingAliases.size(), conflictingAliases);
```
rombert, line 266: Looks like there are three placeholders but only two values.

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
```diff
@@ -36,14 +36,16 @@
  *  Export metrics for the resource resolver bundle:
  *
  *  org.apache.sling.resourceresolver.numberOfResourcesWithAliasesOnStartup -- the total number of resources with sling:alias properties found on startup
+ *  org.apache.sling.resourceresolver.numberOfDetectedConflictingAliases -- the total number of detected invalid aliases
+ *  org.apache.sling.resourceresolver.numberOfDetectedInvalidAliases -- the total number of detected invalid aliases
+ *
  *  org.apache.sling.resourceresolver.numberOfResourcesWithVanityPathsOnStartup -- the total number of resources with sling:vanityPath properties found on startup
- *  org.apache.sling.resourceresolver.numberOfVanityPaths -- the total number of vanity paths in the cache
```
rombert, line 40: This metric name is now missing from the comment, is that intended?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -179,7 +188,12 @@ public MapEntries(final MapConfigurationProvider factory,
         this.aliasMapsMap = new ConcurrentHashMap<>();
         this.stringInterpolationProvider = stringInterpolationProvider;
 
+        this.conflictingAliases = new ConcurrentLinkedQueue<>();
+        this.invalidAliases = new ConcurrentLinkedQueue<>();
```
rombert, line 192: Do `this.conflictingAliases` and `this.invalidAliases` really need to be fields and concurret collections? An approach where they would be used on a single thread and within a single method (or passed around methods) would IMO make it much easier to reason about the concurrency model of this class.

rombert (changes requested):

src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverMetrics.java
reschke: It just moved down; I separated alias from vp metrics, and sorted them alphabetically

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
reschke: It's likely possible but might be more intrusive, because it might affect multiple method signatures. We already have a mix of of fields and parameters, for instance resolveMapsMap.

MapEntries already is convoluted and really needs a refactoring...

Re concurrency: you are right, this doesn't appear to be needed.

reschke: But point taken, please check again now.

rombert (approved):
LGTM, did not look into detail into the refactorings though.

PR #127, opened 2025-01-13
SLING-12611: Resource Resolver: refactor MapEntries - extract PagedQueryIterator

reschke:


rombert (approved):
Makes a lot of sense, thanks.

There will be slight behaviour change with the logger name being changed from ManEntries to PagedQueryIterator. If you think this is a concern feel free to address as part of the same PR, otherwise keep it as is.

reschke:
good point; I considered that as well; will change it

rombert (approved):

reschke:
Sonar is unhappy about PagedQueryIterator to have a cyclic reference because of using the MapEntries logger. Should I just make that public in MapEntries, or pass it down as parameter?

joerghoh:
@reschke I would accept that change in the name of the logger as caused by this refactoring instead of adding additional complexity.

reschke:
(moved utilities related to building JCR-SQL2 queries into separate class; will be extended for aliases)

PR #131, opened 2025-01-17
SLING-12621: Resource Resolver: refactor MapEntries - refactor MapIterator

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntryIterator.java
```diff
@@ -71,69 +78,89 @@ public MapEntry next() {
...
-                this.next = this.nextSpecial;
-                this.nextSpecial = null;
-            }
         } else {
             this.next = this.nextSpecial;
             this.nextSpecial = null;
         }
     }
+
+    private boolean useNextGlobal() {
```
joerghoh, line 126: The conditions in this method look different than above, where you extracted it from. I checked that all 4 return statements were reached by unit tests, so I assume that this reordering of the logic is fine.

joerghoh (changes requested):
In general looks good. I would just harden the handling of the ``specialIterator`` which is not guaranteed to be non-null all the time.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntryIterator.java
```diff
@@ -0,0 +1,166 @@
...
+
+            // given the vanity path in key, walk up the hierarchy until we find
+            // map entries for that path (or stop when root is reached)
+            while (!this.specialIterator.hasNext() && this.key != null) {
+                this.key = removeSelectorsAndExtension(this.key);
+                this.specialIterator = this.getCurrentMapEntryIteratorForVanityPath.apply(this.key);
+                this.key = getParent(key);
+            }
+
+            if (this.specialIterator.hasNext()) {
```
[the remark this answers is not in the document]
reschke: The only way it could be null is when the getCurrentMapEntryIteratorForVanityPath function returns null.

But yes, that's a good idea.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntryIterator.java
reschke: it is; the original code was hard to read and had code duplication.

joerghoh (approved):
+1

PR #132, opened 2025-01-18
Resource Resolver: refactor MapEntries - improve MapIterator test coverage

reschke:
1. Untangle MapIterator from MapEntries for better testability from unit tests
2. Add unit tests

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
@@ -721,7 +721,7 @@ public class MapEntries implements
             key = requestPath.substring(secondIndex);
         }
 
-        return new MapEntryIterator(key, resolveMapsMap.get(GLOBAL_LIST_KEY).iterator(),
+        return new MapEntryIterator(key, resolveMapsMap.get(GLOBAL_LIST_KEY),
                 this::getCurrentMapEntryForVanityPath, this.factory.hasVanityPathPrecedence());
     }
 
@@ -1702,7 +1702,7 @@ public class MapEntries implements
         return l == null ? Collections.emptyIterator() : l.iterator();
     }
 
-    private final class MapEntryIterator implements Iterator<MapEntry> {
+    static final class MapEntryIterator implements Iterator<MapEntry> {
 
         private String key;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1686,12 +1688,19 @@ private void logDisableAliasOptimization(final Exception e) {
                 log.error("A problem occured during initialization of optimize alias resolution. Optimize alias resolution is disabled. Check the logs for the reported problem.", e);
             }
         }
-
     }
 
-    private final class MapEntryIterator implements Iterator<MapEntry> {
+    // return vanity path entry iterator from cache when complete and ready, otherwise from
+    // regular lockup
+    public @Nullable List<MapEntry> getCurrentMapEntryForVanityPath(final String key) {
```
joerghoh, line 1695:
```suggestion
    public @Nullable List<MapEntry> getCurrentMapEntryForVanityPath(@NotNull final String key) {
```

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1704,12 +1713,15 @@ private final class MapEntryIterator implements Iterator<MapEntry> {
         private MapEntry nextSpecial;
 
         private boolean vanityPathPrecedence;
+        private final Function<String, List<MapEntry>> getCurrentMapEntryForVanityPath;
 
-        public MapEntryIterator(final String startKey, final Map<String, List<MapEntry>> resolveMapsMap, final boolean vanityPathPrecedence) {
+        public MapEntryIterator(final String startKey, @NotNull final List<MapEntry> globalList,
```
joerghoh, line 1718: Can you add the ``NotNull`` resp. ``@Nullable`` annotation for all values?

src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -1704,12 +1713,15 @@ private final class MapEntryIterator implements Iterator<MapEntry> {
...
+                                final Function<String, List<MapEntry>> getCurrentMapEntryForVanityPath,
```
joerghoh, line 1719: Right now I don't see the need for the parameter ``getCurrentMapEntryForVanityPath``, as this method is only used in this class. 
Will this change in the future?

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntryIteratorTest.java
```diff
@@ -0,0 +1,158 @@
...
+    public void testExhausted() {
+        assertFalse(empty.hasNext());
+        assertThrows(NoSuchElementException.class,
+                empty::next);
+    }
+
+    @Test
+    public void testRemove() {
+        assertFalse(empty.hasNext());
+        assertThrows(UnsupportedOperationException.class,
```
joerghoh, line 56: just for curiosity: Why do you ``assertThrows`` instead of annotating the ``@Test`` annotation with the expected Exception class?

joerghoh (changes requested):
in general ok, just a few nitpicks; let's add the @Nullable/@NotNull annotations consistently in code which we change.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntryIteratorTest.java
reschke: because Sonar wanted me to :-)

reschke:
The nullability annotations will be added when the class is indeed extracted.

For this PR, I wanted to minimize the changes to what is necessary to improve the test coverage.

joerghoh (approved):
ok, go on

PR #135, opened 2025-01-23
SLING-12634: Resource Resolver: isMaxCachedVanityPathEntriesStartup check does not handle unlimited cache size

reschke:


src/test/java/org/apache/sling/resourceresolver/impl/mapping/VanityPathMapEntriesTest.java
```diff
@@ -1162,26 +1167,28 @@ public void test_getMapEntryList() throws Exception {
                 });
 
         Method method =
-                MapEntries.class.getDeclaredMethod("getMapEntryList",String.class);
+                MapEntries.class.getDeclaredMethod("getMapEntryList", String.class);
         method.setAccessible(true);
         method.invoke(mapEntries, "/target/justVanityPath");
 
+        int expected = 2;
```
joerghoh, line 1174: Nitpick:
```suggestion
        final int expected = 2;
```

joerghoh (approved):
lgtm, you might want to address that single nitpick.

PR #139, opened 2025-02-04
SLING-12649: VanityPathMapEntriesTest - factor out repeated uses of reflection

reschke:


joerghoh (approved):
that makes a lot of sense; wasn't aware that we use that much reflection here.

PR #141, opened 2025-02-11
SLING-12653: add test coverage for query exception disabling optimized alias handling

reschke:


src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
```diff
@@ -222,15 +233,19 @@ private void prepareMapEntriesForAlias(boolean onJcrContent, boolean withNullPar
...
+        when(aliasResource.getValueMap()).thenReturn(buildValueMap(ResourceResolverImpl.PROP_ALIAS, aliases));
 
-        when(resourceResolver.findResources(anyString(), eq("JCR-SQL2"))).thenAnswer((Answer<Iterator<Resource>>) invocation -> {
-            if (invocation.getArguments()[0].toString().contains(ResourceResolverImpl.PROP_ALIAS)) {
-                return List.of(aliasResource).iterator();
-            } else {
-                return Collections.emptyIterator();
-            }
-        });
+        if (queryThrowsWith == null) {
```
rombert, line 238: I think it would be slightly more readable if you would do something like

```java
if ( queryThrowsWith != null ) {
     when(resourceResolver.findResources(anyString(), eq("JCR-SQL2"))).thenThrow(queryThrowsWith);
     return;
}
```

But that's personal preference, feel free to leave as-is.

rombert (approved):

src/test/java/org/apache/sling/resourceresolver/impl/mapping/MapEntriesTest.java
reschke: Yup, different tastes :-)

PR #145, opened 2025-02-18
SLING-12648: MapEntries: factor out vanity path handling into inner class

reschke:
Move into standalone class will be next step; this is supposed to make review easier.


joerghoh (approved):
looks good.

joerghoh:
@reschke you might want to look at the issue presented by Sonar.

reschke:
will check tomorrow

PR #147, opened 2025-02-20
SLING-12665: MapEntries: extract VanityPathHandler into separate class

reschke:


reschke:
Need to rebase first.

PR #149, opened 2025-02-20
SLING-12666: VanityPathHandler: address warnings, refactor

reschke:


reschke:
FWIW, coverage checks fail here, because code was refactored that did not have coverage; will try to address this separately.

reschke:
see #152

PR #151, opened 2025-02-24
SLING-12671: add test coverage for vanity paths on resources with name extensions

reschke:


joerghoh (approved):
extending the testcases makes a lot of sense, thanks for that.

(Although I wonder a bit about the created regexps, but that's a different question.)

PR #152, opened 2025-02-25
SLING-12666: VanityPathHandler, address warnings, refactor

reschke:
1. Renames variables/parameters for clarity
2. fixes whitespace
3. simplifies code
4. extract sub methods when nested too deeply
5. removes the magic number "2" when counting cached map entries for vanity paths
6. other minor Sonar complaints

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/VanityPathHandler.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/VanityPathHandler.java
@@ -375,7 +375,7 @@ public class VanityPathHandler {
      * @param path The resource path to check
      * @return {@code true} if this is valid, {@code false} otherwise
      */
-    boolean isValidVanityPath(final String path){
+    boolean isValidVanityPath(final String path) {
         if (path == null) {
             throw new IllegalArgumentException("Unexpected null path");
         }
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/mapping/VanityPathHandler.java
```diff
@@ -222,30 +223,43 @@ private String getMapEntryRedirect(final MapEntry mapEntry) {
...
-                        }
-                    }
+        String actualContentPath = getActualContentPath(path);
+        List<String> targets = this.vanityTargets.remove(actualContentPath);
+
+        if (targets != null) {
+            for (String target : targets) {
+                int count = removeEntriesFromResolvesMap(target, actualContentPath);
+                if (vanityCounter.longValue() >= count) {
+                    vanityCounter.addAndGet(-count);
```
joerghoh, line 233: why was this changed from ``-2`` to ``-count``? While it sounds logically, is it a bug in the existing implementation?

reschke: There was an assumption throughout the code that the # of cache entries per vanity path always is 2. That is an implementation detail, and I tried - hopefully successfully - to return and use the actual number of additions/removals. Note that this (IMHO) only affects that counter, so it probably does not matter at all.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/VanityPathHandler.java
```diff
@@ -630,23 +630,23 @@ private boolean addEntry(final Map<String, List<MapEntry>> entryMap, final Strin
...
+            entriesCopy.add(entry);
+            Collections.sort(entriesCopy);
+
+            // update map with new list
+            entryMap.put(key, entriesCopy);
+
+            // warn when list of entries for one key grows
+            int size = entriesCopy.size();
+            if (size == 10) {
+                log.debug(">= 10 MapEntries for {} - check your configuration", key);
```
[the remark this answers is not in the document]
reschke: That did not change, but yes, it looks confusing.

I believe the intent was to log *exactly* once when we get to 10 or 100. So maybe the log output should be adjusted.

joerghoh (approved):
lgtm

going through the commits one-by-one helped a lot to understand all the changes, thanks for doing that!

PR #153, opened 2025-02-26
SLING-12679: fix incorrect metrics init

reschke:
`` this.metrics.get().setNumberOfResourcesWithAliasesOnStartupSupplier`` is called twice (first with correct supplier, then with a wrong one)

also sort the counters by (a) category and (b) by name.


joerghoh (approved):
good finding with the incorrect metric.

PR #162, opened 2025-04-01
SLING-12733 Never return null for ResourceMapperImpl.getMapping(...)

kwin:


cziegeler (approved):

joerghoh:
@kwin The javadoc in the interface says:
```
@return a collection of mapped URLs, in no particular order. May not be null or empty.
```

It's not explicitly mandated that it will never return ``null``, but I am fine with this change. But can you please also update the ResourceMapper API description?

PR #164, opened 2025-04-01
remove executable flag on .java

hboutemy:
found while rebuilding release 1.12.6 on Linux that has executable flag support, when release has been done on Windows
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/sling/org.apache.sling.resourceresolver/README.md

just had to chmod a-x...


rombert:
Thanks @hboutemy

PR #165, opened 2025-04-02
 SLING-12734: Move AliasHandler into new class 

reschke:
SonarCloud warnings wrt method complexity will be addressed later on.


joerghoh (approved):
lgtm

thanks for the refactoring

PR #171, opened 2025-04-23
SLING-12758: remove redundant check for absolute path in alias

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -928,20 +928,21 @@ private Resource getChildInternal(final Resource parent, final String childName)
...
                             .map(Map.Entry::getKey);
             if (aliasedResourceName.isPresent()) {
-                final String aliasPath;
-                if (aliasedResourceName.get().startsWith("/")) {
-                    aliasPath = aliasedResourceName.get();
-                } else {
-                    aliasPath = parent.getPath() + '/' + aliasedResourceName.get();
-                }
+                // we know that MapEntries already has checked for valid aliases
+                final String aliasPath = parent.getPath() + '/' + aliasedResourceName.get();
```
joerghoh, line 945: can't you use ``parentPath`` here as well?

PR #173, opened 2025-04-25
SLING-12763: cleanup API of AliasHandler wrt 'optimized resolution'

reschke:
- `AliasHandler.initializeAliases()` returns void now, as the return value was misleading anyway. It has been replaced by `AliasHandler.usesCache()`, indicating the Maps have been initialized.
- Make sure that the `AliasHandler.aliasMapsMap` is always non-null.

This also aligns Alias init with VP init.

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
@@ -67,9 +67,12 @@ class AliasHandler {
     private final Runnable sendChangeEvent;
 
     /**
-     * The key of the map is the parent path, while the value is a map with the  resource name as key and the actual aliases as values
+     * The key of the map is the parent path, while the value is a map with the
+     * resource name as key and the actual aliases as values
      */
     Map<String, Map<String, Collection<String>>> aliasMapsMap;
+
+    boolean mapIsInitialized = false;
 
     final AtomicLong aliasResourcesOnStartup;
     final AtomicLong detectedConflictingAliases;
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
```diff
@@ -130,31 +132,30 @@ protected boolean initializeAliases() {
...
-
-            return isOptimizeAliasResolutionEnabled;
-
         } finally {
 
             this.initializing.unlock();
         }
     }
 
+    boolean usesCache() {
```
joerghoh, line 155: I wonder if this method name is correct, as it returns the value of the variable ``mapsIsInitialized``... (or maybe the name of the variable should be improved?)

src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
```diff
@@ -130,31 +132,30 @@ protected boolean initializeAliases() {
...
                     }
+
                     if (invalidAliases.size() >= MAX_REPORT_DEFUNCT_ALIASES) {
                         log.warn("There are {} invalid aliases; excerpt: {}", invalidAliases.size(), invalidAliases);
                     } else if (!invalidAliases.isEmpty()) {
                         log.warn("There are {} invalid aliases: {}", invalidAliases.size(), invalidAliases);
                     }
                 } catch (final Exception e) {
-
+                    this.aliasMapsMap = new ConcurrentHashMap<>();
```
joerghoh, line 142: That means, that in a case of an exception anywhere in this above code (most likely only in the ``loadAliases`` part) we have zero aliases? That could be considered as a breaking change.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/AliasMapEntriesTest.java
```diff
@@ -1176,7 +1177,8 @@ public void test_doRemoveAliasFromSibling() throws Exception {
     public void test_initAliasesAfterDispose() {
         AliasHandler ah = mapEntries.ah;
         mapEntries.dispose();
-        boolean enabled = ah.initializeAliases();
-        assertFalse("return value (isOptimizeAliasResolutionEnabled) should be false", enabled);
+        ah.initializeAliases();
+        boolean enabled = ah.usesCache();
+        assertFalse("alias handler should not have set up cache", enabled);
```
joerghoh, line 1182:
```suggestion
        assertFalse("alias handler should not have set up cache", ah.usesCache());
```

src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
reschke: Hm, no. Previously we wouldn't even have used AliasHandler because of "useOptimized..." being false.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
joerghoh: better 👍

src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
joerghoh: You are correct. While before in the case of an exception ``this.aliasMapsMap`` was left with whatever was already stored there, ``mapEntries.isOptimizedAliasResolutionEnabled()``was returning false in this case and the ``aliasMapsMap`` was never consulted.

For that this is just a resource optimizations, and you could achieve the same with just a clearing of this map.
```suggestion
                    // Get rid of whatever is already stored here, it will never get used again
                    this.aliasMapsMap.clear();
```

reschke: Actually, I'll make that an unmodifiable map for now (no additional review needed :-).

PR #174, opened 2025-04-29
 SLING-12769: alias refactoring - MapEntries - move logic related to alias events into AliasHandler

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
```diff
@@ -159,7 +159,11 @@ boolean usesCache() {
     }
 
     boolean doAddAlias(final Resource resource) {
-        return loadAlias(resource, this.aliasMapsMap, null, null);
+        if (cacheIsInitialized) {
```
joerghoh, line 162: right now ``cacheIsInitialized`` is not marked as volatile. But with this change (or maybe even before) we should do that, as it can be accessed by multiple threads (at least that's my assumption).

joerghoh (approved):
lgtm

PR #175, opened 2025-05-06
SLING-12777: alias refactoring - make ResourceMapperImpl ignorant of optimimized/non-optimized case

reschke:
This removes logic from `ResourceMapperImpl` that previously took two different code paths, depending on the mode.

To get there, `AliasHandler` now handles `getAliasMap` even when the alias cache is not in use.

Note: for that, it uses its own resolver instance, thus the test in `ResourceMapperImplTest` does not "work" anymore (code removed for now). Suggestions how to resurrect that check (do we even need it anymore?) are welcome.

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
@@ -343,7 +343,7 @@ class AliasHandler {
                 Resource parent = resolver.getResource(parentPath);
                 return getAliasMapFromRepo(parent);
             } catch (LoginException ex) {
-                log.error("Could not obtain resolver", ex);
+                log.error("Could not obtain resolver to resolve any aliases from repository", ex);
                 return null;
             }
         }
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
@@ -21,6 +21,7 @@ package org.apache.sling.resourceresolver.impl.mapping;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
```diff
@@ -308,14 +311,52 @@ private boolean doUpdateAliasInMap(final Resource resource) {
...
+            Resource parent = resolver.getResource(parentPath);
+            result = getAliasMapFromRepo(parent);
+        } catch (LoginException ex) {
+            log.error("Could not obtain resolver", ex);
+        }
+
+        return result;
+    }
+
+    private @Nullable Map<String, Collection<String>> getAliasMapFromRepo(final Resource parent) {
```
joerghoh, line 346:
```suggestion
    private @Nullable Map<String, Collection<String>> getAliasMapFromRepo(@Nullable final Resource parent) {
```

joerghoh (approved):
LGTM ... maybe a few more @Nullable/@NotNull annotation could be helpful.

src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
```diff
@@ -300,40 +308,80 @@ private boolean doUpdateAliasInMap(final Resource resource) {
...
+        if (parentPath == null) {
+            return null;
+        } else {
+            try (ResourceResolver resolver =
+                    factory.getServiceResourceResolver(factory.getServiceUserAuthenticationInfo(SERVICE_USER))) {
+
+                Resource parent = resolver.getResource(parentPath);
+                return getAliasMapFromRepo(parent);
+            } catch (LoginException ex) {
+                log.error("Could not obtain resolver", ex);
```
joerghoh, line 346: I would make this log message a bit more helpful
```suggestion
                log.error("Could not obtain resolver to resolve any aliases from repository", ex);
```

joerghoh (approved):
+1 (besides that single suggestion to improve the exception message ...)

PR #177, opened 2025-05-14
SLING-12759: ResourceResolver: push 'non-optimized' alias handling from ResourceResolverImpl into AliasHandler

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -930,7 +930,7 @@ private Resource getChildInternal(final Resource parent, final String childName)
         final String parentPath = parent.getPath();
         logger.debug("getChildInternal: looking up {} in {}", childName, parentPath);
 
-        final Optional<String> aliasedResourceName = factory.getMapEntries().getAliasMap(parentPath).entrySet().stream()
+        final Optional<String> aliasedResourceName = factory.getMapEntries().getAliasMap(parent).entrySet().stream()
```
joerghoh, line 933: Nit: when you don't need the ``parentPath`` here but only in the branch below (line 938ff), I would also move its declaration down there.

joerghoh (approved):
lgtm -- you might want to address the 2nd nit

PR #178, opened 2025-05-19
SLING-12787: ResourceResolver: alias refactoring - use Resource instead of Path when getting aliases

reschke:
This changes `ResourceMapperImpl.resolveAliases` to use the `Resource` that it already has for (a) getting the aliases and (b) walking the tree up.

Changes in tests unfortunately were needed as:

- resources were mocked without handling `getParent()`, and
- decorator tests were using inconsistent mocked values for `getParent()` and `getName()`

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
@@ -83,12 +83,6 @@ public class ResourceResolverImpl extends SlingAdaptable implements ResourceReso
 
     public static final String PROP_ALIAS = "sling:alias";
 
-    // The suffix of a resource being a content node of some parent
-    // such as nt:file. The slash is included to prevent false
-    // positives for the String.endsWith check for names like
-    // "xyzjcr:content"
-    public static final String JCR_CONTENT_LEAF = "/jcr:content";
-
     protected static final String PARENT_RT_CACHEKEY = ResourceResolverImpl.class.getName() + ".PARENT_RT";
 
     /** The factory which created this resource resolver. */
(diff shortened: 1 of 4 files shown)


reschke:
ok,

with the only diff from master...

```
diff --git a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
index 2366ebc..b314df5 100644
--- a/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/mapping/ResourceMapperImpl.java
@@ -233,8 +233,9 @@ public class ResourceMapperImpl implements ResourceMapper {

     private void resolveAliases(Resource res, PathGenerator pathBuilder) {
         String path = res.getPath();
+        Resource current = res;

-        while (path != null) {
+        while (path != null && current != null) {
             Collection<String> aliases = Collections.emptyList();
             // read alias only if we can read the resources and it's not a jcr:content leaf
             if (!path.endsWith(ResourceResolverImpl.JCR_CONTENT_LEAF)) {
@@ -243,9 +244,15 @@ public class ResourceMapperImpl implements ResourceMapper {
             // build the path from the name segments or aliases
             pathBuilder.insertSegment(aliases, ResourceUtil.getName(path));
             path = ResourceUtil.getParent(path);
+
             if ("/".equals(path)) {
                 path = null;
+            } else if (current.getParent() == null) {
+                throw new RuntimeException("current.getParent() == null for " + current.getClass()
+                        + ", while current.getPath() == " + current.getPath());
             }
+
+            current = current.getParent();
         }
     }
```

I get...
```
[ERROR] Errors:
[ERROR]   MockedResourceResolverImplTest.testMapping:660 » Runtime current.getParent() == null for class org.apache.sling.api.resource.Resource$MockitoMock$tZqPZmm6, while current.getPath() == /single/test
```

So it seems the mocked resource is broken - but it's generated in sling.api.

joerghoh (approved):
lgtm

PR #182, opened 2025-05-28
SLING-12806 Prevent creating a resource with a name only consisting out

kwin:
of dots


reschke (commented):
Is "..." really illegal?

cziegeler (approved):

PR #184, opened 2025-06-05
SLING-12823: alias refactoring - avoid getting the parent resource when path sufficient

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/AliasHandler.java
```diff
@@ -469,27 +469,17 @@ private boolean loadAlias(
...
             return false;
         } else {
-            Resource parent = containingResource.getParent();
-
-            if (parent == null) {
-                log.warn(
-                        "{} is null for alias on {}, skipping.",
-                        containingResource == resource ? "parent" : "grandparent",
-                        resource.getPath());
+            String[] aliasArray = resource.getValueMap().get(ResourceResolverImpl.PROP_ALIAS, String[].class);
```
joerghoh, line 472: I wonder why you use ``resource``here. Shouldn't it be ``containingResource``?

reschke: it's correct.

containingResource != resource if and only if resource is a jcr:content resource.

In which case we read the aliases on jcr:content, and apply them to the container.

(and yes, alignment with the cached case is coming as well, that's a different PR)

Will add some comments so this becomes clearer.

reschke: See https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/184/commits/8c2cc91f037878c80bd612e4a83791242020523b

PR #190, opened 2025-08-14
 SLING-12888: Parametrize AliasMapEntriesTest for optimized/non-optimized testing

reschke:
This parametrizes the test class to run with "optimized" on/off.

Several tests only make sense when running in "optimized" node, thus they are skipped to the other mode.

Also some more cleanup.


reschke:
found flaky test, checking...

PR #193, opened 2025-08-21
SLING-12900: improve test coverage for VP bg init vs events, minor refactoring

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/MapEntries.java
```diff
@@ -450,28 +454,27 @@ public void onChange(final List<ResourceChange> changes) {
...
-            // during startup: just enqueue the events
-            if (inStartup) {
-                if (type == ResourceChange.ChangeType.REMOVED
-                        || type == ResourceChange.ChangeType.ADDED
-                        || type == ResourceChange.ChangeType.CHANGED) {
+            boolean queued = false;
+
+            if (RELEVANT_CHANGE_TYPES.contains(type)) {
+                // during startup: just enqueue the events
+                if (inStartup) {
```
joerghoh, line 468: can you merge these 2 conditions (and probably change the order, as the boolean check is faster)?

joerghoh (approved):
+1 in general, but just a small nitpick in there (sonar will probably complain about it as well).

PR #195, opened 2025-08-22
SLING-12894: alias refactoring - support observation events while bg init not finished

reschke:


reschke:
See e57c136781d57bd46e8369aab05408d1694878b4

PR #200, opened 2025-08-26
SLING-12912: alias handling: wire alias bg cache init to OSGi

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverFactoryConfig.java
```diff
@@ -168,10 +168,15 @@ int resource_resolver_default_vanity_redirect_status() default
...
             description = "This flag controls whether to optimize"
-                    + " the alias resolution by creating an internal cache of aliases. NOTE: this setting will be removed "
-                    + " with the optimized resolution being the default (see SLING-12025).")
+                    + " the alias resolution by creating an internal cache of aliases.")
     boolean resource_resolver_optimize_alias_resolution() default true;
 
+    @AttributeDefinition(
+            name = "Alias Cache Init In Background",
+            description =
+                    "This flag controls whether the vanity path cache will be initialized immediately or later as a background task.")
```
joerghoh, line 177:
```suggestion
                    "This flag controls whether the vanity path cache will be initialized synchronously or as a background thread (false), not blocking the startup (true).")
```

joerghoh (commented):
let's the have documentation a bit clearer :-)

PR #205, opened 2026-05-20
SLING-13206: ResourceDecorator services lost after @Modified on ResourceResolverFactoryActivator

royteeuwen:
When OSGi fires the @Modified lifecycle callback (e.g. triggered by a bundle deployment causing a configuration change), modified() delegates to deactivateInternal() which calls resourceDecoratorTracker.close(), clearing all registered ResourceDecorator services.

Because the decorators use ReferencePolicy.DYNAMIC, OSGi does not re-fire bindResourceDecorator after a @Modified event (no component restart). The tracker is therefore left permanently empty, silently breaking any feature that relies on ResourceDecorator (e.g. path-rewriting decorators).

Fix: introduce deactivateInternalWithoutClearingDecorators() which performs the same teardown as deactivateInternal() but omits the tracker.close() call. @Modified now delegates to this new helper. @Deactivate continues to use the full deactivateInternal() where a complete teardown is correct.

Also add a null-guard in ResourceDecoratorEntry.compareTo() to prevent NPE when ServiceReference is null.

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/helper/ResourceDecoratorTracker.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/helper/ResourceDecoratorTracker.java
@@ -141,6 +141,15 @@ public class ResourceDecoratorTracker {
         }
 
         public int compareTo(final ResourceDecoratorEntry o) {
+            if (this.comparable == null && o.comparable == null) {
+                return 0;
+            }
+            if (this.comparable == null) {
+                return -1;
+            }
+            if (o.comparable == null) {
+                return 1;
+            }
             return comparable.compareTo(o.comparable);
         }
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/helper/ResourceDecoratorTracker.java
```diff
@@ -141,6 +141,9 @@ public ResourceDecoratorEntry(final ResourceDecorator d, final Comparable<Object
         }
 
         public int compareTo(final ResourceDecoratorEntry o) {
+            if (this.comparable == null || o.comparable == null) {
```
cziegeler: while this prevents NPEs, it is not 100% correct, if one of them is null but the other is not

royteeuwen: Fixed

cziegeler (approved):

PR #211, opened 2026-07-07
SLING-13260 adding CharSet to instantiate the String

pat-lego:
Adding an enhancment to prevent locking to instantiate a String.

### DIFF
--- a/src/main/java/org/apache/sling/resourceresolver/impl/helper/URI.java
+++ b/src/main/java/org/apache/sling/resourceresolver/impl/helper/URI.java
@@ -24,6 +24,7 @@ import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.BitSet;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/sling/resourceresolver/impl/helper/URI.java
```diff
@@ -4018,11 +4019,7 @@ private static String getAsciiString(final byte[] data) {
             throw new IllegalArgumentException("Parameter may not be null");
         }
 
-        try {
-            return new String(data, "US-ASCII");
-        } catch (UnsupportedEncodingException e) {
-            throw new URIException("HttpClient requires ASCII support");
-        }
+        return new String(data, StandardCharsets.US_ASCII);
```
reschke, line 4022: what happens if the byte array does not map to ASCII?

reschke (approved):

src/main/java/org/apache/sling/resourceresolver/impl/helper/URI.java
reschke, line 4070: Such as this one?

reschke (commented):
https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#String-byte:A-java.nio.charset.Charset-

So this should be ok if data loss is no problem

src/main/java/org/apache/sling/resourceresolver/impl/helper/URI.java
pat-lego: This function assumes it is Ascii (getAsciiString) it would be assumed that it should be otherwise why are we calling this function.

PR #215, opened 2026-07-17
SLING-13265: Wrong ResourceMetadata for alias resolution with unreadable ancestor

jsedding:


src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java
```diff
@@ -809,59 +809,57 @@ private static String getMapPath(final String scheme, final String host, int por
...
 
-            while (resource != null && tokener.hasMoreTokens()) {
+            // we found an ancestor resource, so we need to prefix the resolutionPath with it
+            // unless it is the root path (that would result in a double slash prefix "//")
+            String resolutionPathPrefix = resource.getResourceMetadata().getResolutionPath();
+            if (!Objects.equals(resolutionPathPrefix, "/")) {
+                resolutionPath.append(resolutionPathPrefix);
+            }
+
+            while (tokener.hasMoreTokens()) {
```
reschke: Is the removal of the `null` check intentional?

Later on, `resource` is passed to `getChildInternal`, where it might be dereferenced.

reschke: (tried to write a test but failed so far)

jsedding: good catch, that was a regression 🙂 it's fixed now and covered with a test.

PR #224, opened 2026-07-31
SLING-13284 - Do not abort alias/vanity cache init on out-of-order qu…

sagarmiglani:
…ery results

The ordering check in PagedQueryIterator compared live property values (read from the repository) against the sort order delivered by an async Lucene index. Since async indexes are eventually consistent, their sort order can lag behind the current repository state after a property write. The previous behavior treated this lag as a fatal error and aborted the entire alias/vanity cache build, leaving alias resolution non-functional until a manual reindex + restart.

Demote the two throws to log.warn and continue processing. The row contains a valid alias/vanity value that should be cached regardless of sort position. This also addresses the vanity-path variant (SLING-13248).


reschke (commented):
We should do this in master branch first.

reschke (approved):

sagarmiglani:
closing this in favour of https://github.com/apache/sling-org-apache-sling-resourceresolver/pull/231

PR #226, opened 2026-08-04
SLING-13287: paging sanity checks - log type when not String

reschke:
Note that the test does not verify the log output (which makes it kind of lame). Do we have a simple way to do this, such as LogCustomizer in Oak?

### DIFF
--- a/src/test/java/org/apache/sling/resourceresolver/impl/mapping/PagedQueryIteratorTest.java
+++ b/src/test/java/org/apache/sling/resourceresolver/impl/mapping/PagedQueryIteratorTest.java
@@ -111,8 +111,9 @@ public class PagedQueryIteratorTest extends AbstractMappingMapEntriesTest {
 
     @Test
     public void testSimpleWrongType() {
-        try (TestLogger logger =
-                TestLogger.createStartedFor(PagedQueryIterator.class).contains("unexpected")) {
+        try (TestLogger logger = TestLogger.create(PagedQueryIterator.class)
+                .contains("unexpected")
+                .start()) {
 
             String[] expected = new String[] {"a", "b", "c"};
             Collection<Resource> expectedResources = toResourceList(expected);
@@ -291,12 +292,12 @@ public class PagedQueryIteratorTest extends AbstractMappingMapEntriesTest {
 
         private final Appender<ILoggingEvent> customLogger;
 
-        private final Logger logger;
+        private final Logger delegate;
         private String matchContainsMessage;
         private final List<String> logs = Collections.synchronizedList(new ArrayList<>());
 
         private TestLogger(Class<?> clazz) {
-            this.logger = getLogger(clazz);
+            this.delegate = getLogger(clazz);
 
             this.customLogger = new AppenderBase<>() {
                 @Override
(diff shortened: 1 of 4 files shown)


src/test/java/org/apache/sling/resourceresolver/impl/mapping/PagedQueryIteratorTest.java
```diff
@@ -98,14 +108,49 @@ public void testSimpleWrongOrder() {
...
+            expectedResources.add(r);
+
+            when(resourceResolver.findResources(eq("testSimpleWrongType"), eq("JCR-SQL2")))
+                    .thenReturn(expectedResources.iterator());
+            PagedQueryIterator it =
+                    new PagedQueryIterator("alias", PROPNAME, resourceResolver, "testSimpleWrongType", 2000);
+
+            String[] expWithOneMore = Arrays.copyOf(expected, expected.length + 1);
+            // implementation detail: this assumes the way Sling converts Dates to Strings
+            expWithOneMore[expected.length] = oneMore.toInstant().toString();
```
jsedding: Why not use the conversion that _is implemented_ in Sling?

```suggestion
            expWithOneMore[expected.length] = properties.get(PROPNAME, new String[] {"dummy"})[0];
```

jsedding (approved):
Looks good. Just a minor suggestion for the unit-test.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/PagedQueryIteratorTest.java
```diff
@@ -98,14 +108,49 @@ public void testSimpleWrongOrder() {
...
+    @Test
+    public void testSimpleWrongType() {
+        try (TestLogger logger =
+                TestLogger.createStartedFor(PagedQueryIterator.class).contains("unexpected")) {
+
+            String[] expected = new String[] {"a", "b", "c"};
+            Collection<Resource> expectedResources = toResourceList(expected);
+
+            Date oneMore = new Date(0);
+            ValueMap properties = new ValueMapDecorator(Map.of(PROPNAME, new Date[] {oneMore}));
```
joerghoh, line 116: I don't think if this behavior is identical to the behavior of the JcrValueMap, when it converts the underlying JCR property types to the request types...

To be on the safe side, I would start with a Sling Mock backed by a JCR OAK repository; that could make the test more complex (and require more dependencies), but then you can have the identical behaviour.

src/test/java/org/apache/sling/resourceresolver/impl/mapping/PagedQueryIteratorTest.java
reschke: Not sure how this would work. What I need here is Sling's String representation `Date(0` - how can I get that?

src/test/java/org/apache/sling/resourceresolver/impl/mapping/PagedQueryIteratorTest.java
reschke: Currently, Resource Resolver has no Jackrabbit dependencies. AFAIK.

PR #228, opened 2026-08-12
SLING-13301: test syntax variants

reschke:


src/test/java/org/apache/sling/resourceresolver/impl/mapping/VanityPathMapEntriesTest.java
```diff
@@ -267,6 +267,66 @@ public void test_simple_vanity_path() {
...
+        internal_test_syntax_variants("//x//y", "//x//y");
+        // multi-segment absolute path starting with "/.."
+        internal_test_syntax_variants("/../x", "/../x");
+        // blank path
+        internal_test_syntax_variants(" ", null);
+        // empty path
+        internal_test_syntax_variants("", null);
+    }
+
+    private void internal_test_syntax_variants(String vanityPath, String expectation) {
```
joerghoh:
```suggestion
    private void assertPathResolvesTo(String vanityPath, String expectedPath) {
```

joerghoh (commented):
NIT: that method name could be a bit more expressive :-)

PR #234, opened 2026-08-27
SLING-13323: when paging encounters unexpected values, log the position inside the page as well

reschke:


src/main/java/org/apache/sling/resourceresolver/impl/mapping/PagedQueryIterator.java
```diff
@@ -104,8 +107,9 @@ private Resource getNext() throws NoSuchElementException {
             String value = values[0];
             if (value.compareTo(lastKey) < 0) {
                 log.warn(
-                        "unexpected query result in page {}, property name '{}', got '{}'{}, despite querying for > '{}'"
+                        "unexpected query result at index {} in page {}, property name '{}', got '{}'{}, despite querying for > '{}'"
```
sagarmiglani: Could we just pass count directly into the two  log.warn(...)  calls instead of adding a new field?

joerghoh (approved):

